PR #1, opened 2021-07-29
More improvements

cbrnr:
I've further refined our README.md. Now I'll go through the rest of the code one last time.

@hofaflo is it OK if we spell our project SleepECG?


cbrnr:
I think we're ready for the first release. Let me know what you think, if this is merged we could go for it. Can you quickly summarize the steps required for a release? Is creating a release on GitHub enough?

README.md
```diff
@@ -1,47 +1,48 @@
...
+For analysis of heartrate variability, detecting the exact location of heartbeats is essential. As a measure of how accurately a detector is, we computed Pearson's correlation coefficient between resampled RRI time series deduced from annotated and detected beat locations from all [GUDB](https://github.com/berndporr/ECG-GUDB) records. Our implementation detects peaks in the bandpass-filtered ECG signal, so it produces
+stable RRI time series without any post-processing.
```
hofaflo, line 40:
```suggestion
For analysis of heartrate variability, detecting the exact location of heartbeats is essential. As a measure of how accurate a detector is, we computed Pearson's correlation coefficient between resampled RRI time series deduced from annotated and detected beat locations from all [GUDB](https://github.com/berndporr/ECG-GUDB) records. Our implementation detects peaks in the bandpass-filtered ECG signal, so it produces stable RRI time series without any post-processing.
```

hofaflo (changes requested):
Fixed a typo and stray newline.

hofaflo:
TODO for release:
- [ ] change pypi-login data in [`.github/workflows/release.yml`](https://github.com/cbrnr/sleepecg/blob/3b274ce20aeec0e34b088eecd1c10b7511a950a0/.github/workflows/release.yml#L80-L81) to use your stored secrets
- [ ] create a GitHub release with tag `v0.1.0`

cbrnr:
I added the API token to the repository secrets, so I think this should work.

hofaflo:
Perfect, then let's give it a try!

cbrnr:
OK with adding the CHANGELOG.md?

cbrnr:
One more thing: we should have a `sleepecg.__version__` attribute. Currently, we store the version only in `setup.cfg` AFAIK. Would you just duplicate this information by simply adding `__version__ = '0.1.0'` or do you want to try to read it from the config file? I tend to prefer the first option because it seems like it's not that much effort.

cbrnr:
Seems like there's a nice option: https://packaging.python.org/guides/single-sourcing-package-version/ (the Note section in the first bullet point). I'll give that a try.

hofaflo:
> Seems like there's a nice option: https://packaging.python.org/guides/single-sourcing-package-version/ (the Note section in the first bullet point). I'll give that a try.

Yep, just tried that, should work!

PR #3, opened 2021-07-29
Fix MemoryError in case of invalid checksum

hofaflo:
`io.utils.download_file` calls itself to retry the download in case checksum verification fails. In case the "true" checksum is incorrect, this leads to an endless recursion stopped eventually by a `MemoryError`. I suggest to raise a `RuntimeError` instead of retrying. Alternatively, we could introduce a parameter to specify how many times the download should be retried.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## [UNRELEASED]
+### Fixed
+- Fix MemoryError in case of invalid checksum ([#3](https://github.com/cbrnr/sleepecg/pull/3))
 
 ## [0.1.0] - 2021-07-29
 ### Added
--- a/sleepecg/io/utils.py
+++ b/sleepecg/io/utils.py
@@ -95,5 +95,5 @@ def download_file(
             raise RuntimeError(
                 f'Checksum mismatch for {target_filepath}:\n'
                 f'    {checksum!r} (expected)\n'
-                f'    {calculated_checksum!r} (calcuated)',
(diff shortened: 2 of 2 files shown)


cbrnr:
Good idea! Some suggestions:

1. I'm fine with raising an exception, no need to retry.
2. I wouldn't include the `Raises` section here because the error message should be clear already (see also the recommendation on this optional section [in the docs](https://numpydoc.readthedocs.io/en/latest/format.html#raises)).
3. I'd include the actual and expected checksums in the error message – unless of course the string is too long. WDYT?
4. Can you add an entry to `CHANGELOG.md` (I guess under "Fixes")?

hofaflo:
> 2. I wouldn't include the `Raises` section here because the error message should be clear already (see also the recommendation on this optional section [in the docs](https://numpydoc.readthedocs.io/en/latest/format.html#raises)).

Okay, I included a sentence in the function description, so one doesn't need to try out how it will crash in case verification fails.

cbrnr:
Can you show how an error would look like?

hofaflo:
```
RuntimeError: Checksum mismatch for README.md:
    '06e13a3e2bd156e66c49fdbc10bc78be' (excpected)
    '06e13a3e2bd156e66c49fdbc10bc78bf' (calcuated)
```

cbrnr:
Another typo: `calcuated` -> `calculated` 😄

cbrnr:
Thanks @hofaflo!

PR #4, opened 2021-07-29
Faster build+test workflow

hofaflo:
The build+test workflow (in `test.yml`) is triggered on PRs. Currently, it builds wheels for cp37, cp38 and cp39 for each platform. To shorten workflow runtime (and save runner minutes), I'd suggest the following steps:
- [x] cancel previous in-progress workflows using concurrency groups
- [ ] ~~build only cp37 (the minimum required python version) wheels (but keep building+testing all wheels in the release-workflow)~~
- [ ] use cached pip package files (not sure how this works with cibuildwheel yet)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,11 +33,11 @@ jobs:
       CIBW_TEST_COMMAND: pytest {package}
       CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
       CIBW_MANYLINUX_I686_IMAGE: manylinux2014
-      CIBW_SKIP: cp37*
+      CIBW_BUILD: cp37*
       CIBW_ARCHS_MACOS: x86_64 arm64
 
     needs: [pre-commit]
-    name: Build wheels on ${{ matrix.os }}
+    name: Build cp37 wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
 
     strategy:
@@ -67,7 +67,7 @@ jobs:
       CIBW_ARCHS_MACOS: x86_64 arm64
 
     needs: [pre-commit]
-    name: Build wheels on ${{ matrix.os }}
+    name: Build all wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
 
     strategy:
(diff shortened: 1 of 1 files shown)


cbrnr:
We have to create a release before we can test if all wheels build correctly then. I'm fine with only building cp37 wheels by default, but can we add an option to manually trigger the builds for all versions? In MNE-Python, we do this by prepending a specific tag (e.g. `[SKIPCI]`) to the commit message. If we decide to implement this, we need to make sure to document it (e.g. in `CONTRIBUTING.md` where I'd also want to put instructions on how to make a release) – otherwise I'm sure I will forget 😄 !

If you think this is not necessary LMK, maybe I'm overly cautious.

cbrnr:
Maybe let's just use the concurrency change and continue to build all wheels? It doesn't seem to be too slow at the moment.

cbrnr:
I see the commits from the previous (already merged) PRs – could you rebase onto `main` first and then force-push?

hofaflo:
To summarize the above trial- and error process:
- one could use two different jobs for building wheels and decide between them using an if-clause
- in the past, it was possible to get the last commit message like this: `if: "!contains(github.event.head_commit.message, 'ci skip')"`, which could then be checked for a substring (e.g. "[buildall]" to trigger building wheels for all versions)... apparently this has been removed and I couldn't find another way to access the message
- it's possible use e.g. "[skip ci]" in a commit message (see [here](https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/)), but this seems to disable CI for the entire PR, so not a good option either

So, if you're not worried about your runner minutes, let's leave it at the concurrency change. Otherwise we could also create an additional workflow for building all wheels which can be triggered manually and on release creation (instead of publication)?

cbrnr:
I'm not worried about CI time at the moment, so let's just merge the concurrency change for now.

hofaflo:
Okay great! 74643de includes concurrency and your change from #7. If you want to avoid the trial- and error history, merging 
b768ea9 should also work.

Side note: Checks are running again, so including `[skip ci]` in the commit message looks like the way to go e.g. for typos or changes in the readme - but no idea why they didn't run for c04f8d9 and 64b4b47 :/

cbrnr:
Great! It looks like this PR only contains the concurrency change and nothing else, so rebasing worked! I'll squash and merge anyway, so the trial and error is OK. Thanks @hofaflo!

PR #8, opened 2021-08-01
Add interface to download NSRR data

hofaflo:
The National Sleep Research Resource (NSRR) offers access to sleep datasets. A [ruby gem](https://github.com/nsrr/nsrr-gem) and an [R package](https://github.com/muschellij2/nsrr) are available, but I couldn't find a python interface.


cbrnr:
Thanks @hofaflo!

PR #9, opened 2021-08-03
Fix incorrect names in heartbeat_detection

hofaflo:
Found two naming mistakes:
- `heartbeat_detection.py:95` mentions `scipy.signal.sosfilt`, although `sosfilfilt` is used (and causes the described issue)
- in `_heartbeat_detection.c:_squared_moving_integration` the pointer to the output array's data is called `beat_mask` (copy-paste error from `_thresholding`)


cbrnr:
Thanks @hofaflo!

PR #12, opened 2021-08-08
Fix naming inconsistencies in readme code listing

hofaflo:
The array containing the detected peak indices is called `detection` for all detectors but the two from neurokit.


cbrnr:
Thanks @hofaflo!

PR #18, opened 2021-08-16
Shorten names

cbrnr:
Some of the names are too verbose and could be shortened without losing their descriptiveness.

PR #20, opened 2021-08-17
Make NamedTuples used as returnvalues non-public

hofaflo:
They aren't required in the public API and this keeps the autogenerated documentation clean.


cbrnr:
I wonder how this renders in the docs, because we do need the type as the return value for a public function...

hofaflo:
Since each element in the tuple is documented individually, it's not an issue:
image

hofaflo:
rebased so renaming `heartbeat_detection.py`->`heartbeats.py` from #18 is considered, ready to merge!

cbrnr:
Thanks @hofaflo!

PR #21, opened 2021-08-17
Standardize return value docs

hofaflo:
I'd suggest we follow the [pandas docstring guide](https://pandas.pydata.org/pandas-docs/stable/development/contributing_docstring.html#section-4-returns-or-yields) for documenting return values, since it's more specific than [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#returns) in that matter;
- for single return values, only the type is stated (no name)
- for multiple return values, both a name and a type are stated
- generators are treated similarly, so a type annotation `Iterator[int]` becomes just `int` in the docstring

### DIFF
--- a/sleepecg/heartbeats.py
+++ b/sleepecg/heartbeats.py
@@ -86,7 +86,7 @@ def detect_heartbeats(ecg: np.ndarray, fs: float, backend: str = 'c') -> np.ndar
 
     Returns
     -------
-    heartbeat_indices: np.ndarray
+    np.ndarray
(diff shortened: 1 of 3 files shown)


cbrnr:
Nice! The pandas guidelines make sense, because I was already wondering if we can/should omit the name when multiple values (in a tuple) are returned. Thanks @hofaflo!

cbrnr:
BTW, this should go into our upcoming `CONTRIBUTING.md` document.

Issue #6, opened 2021-07-30
conda-forge package


hoechenberger:
Hello @cbrnr, congrats to the public release!

Are you planning to have a package on `conda-forge` too? I can help you create it (or create it for you ;)). Just let me know!

cbrnr:
Thanks @hoechenberger, if you have time to do it please go ahead! We can include the conda install command in our README once it's available.

hoechenberger:
WIP:
https://github.com/conda-forge/staged-recipes/pull/15932

hoechenberger:
Just FYI, the recipe has been merged, but there's currently a CI problem that prevents the packages from being built & deployed. Hopefully this will be resolved soon.

cbrnr:
Thanks @hoechenberger! Let us know when it is available and we can include a link in our `README.md`. Also, we will release 0.3 soon so you don't have to vendor `LICENSE` anymore (I guess this will require an update to the formula?).

hoechenberger:
> Also, we will release 0.3 soon so you don't have to vendor `LICENSE` anymore (I guess this will require an update to the formula?).

That will be just a really tiny change, I'll take care of this once 0.3 has been released :)

hoechenberger:
The package is now available from conda-forge:

https://anaconda.org/conda-forge/sleepecg

Before doing an official announcement, we should probably ensure it actually works as intended 😅 I haven't tested the various "backends"; my understanding is that there's Python, numba, and C, right? They all SHOULD work but would be good to check

cbrnr:
Nice! You can test them with our tests. Do you have `numba` as a dependency or is it just optional? In any case, we fall back to pure Python if neither C nor Numba are available (so the order of preference is C -> Numba -> Python).

hoechenberger:
I included all optimal dependencies 😅 and linked the C module against NumPy
Are the tests included in the sdist?

I'm out for coffee, will have a look at this later today 💪

hofaflo:
Just ran the tests, seems to work, thanks @hoechenberger! :)

cbrnr:
Tests are not included in the sdist. It seems like we should just create a `MANIFEST.in` containing `*`.

hoechenberger:
@cbrnr sounds fair!

@hofaflo awesome, thanks! I think it's safe to announce the conda-forge package then!

cbrnr:
Thanks @hoechenberger! If you have any idea how we can increase visibility, let us know at #25!

PR #22, opened 2021-08-18
Add code used for benchmarks as example

hofaflo:


README.md
```diff
@@ -29,7 +29,9 @@ detection = detect_heartbeats(ecg, fs)
 
 
 ### Performance evaluation
-We evaluated detector runtime using slices of different lengths from [LTDB](https://physionet.org/content/ltdb/1.0.0/) records with at least 20 hours duration. Error bars in the plot below correspond to the standard error of the mean. Our detector (we only show the C backend) is by far the fastest implementation among all tested packages (note that the *y*-axis is logarithmically scaled).
+All code used for performance evaluation is available under [`examples/benchmark`](https://github.com/cbrnr/sleepecg/tree/main/examples/benchmark). The used package versions are listed in [`requirements-benchmark.txt`](https://github.com/cbrnr/sleepecg/blob/main/examples/benchmark/requirements-benchmark.py)
```
cbrnr, line 32:
```suggestion
All code used for performance evaluation is available in [`examples/benchmark`](https://github.com/cbrnr/sleepecg/tree/main/examples/benchmark). The used package versions are listed in [`requirements-benchmark.txt`](https://github.com/cbrnr/sleepecg/blob/main/examples/benchmark/requirements-benchmark.py)
```

README.md
```diff
@@ -29,7 +29,9 @@ detection = detect_heartbeats(ecg, fs)
...
+
+We evaluated detector runtime using slices of different lengths from [LTDB](https://physionet.org/content/ltdb/1.0.0/) records with at least 20 hours duration. Error bars in the plot below correspond to the standard error of the mean. The C backend of our detector is by far the fastest implementation among all tested packages (note that the *y*-axis is logarithmically scaled). Runtime evaluation was performed on an [Intel® Xeon® Prozessor E5-2440 v2](https://ark.intel.com/content/www/us/en/ark/products/75263/intel-xeon-processor-e5-2440-v2-20m-cache-1-90-ghz.html) and 31.3 GiB RAM. No parallelization was used.
```
cbrnr, line 34:
```suggestion
We evaluated detector runtime using slices of different lengths from [LTDB](https://physionet.org/content/ltdb/1.0.0/) records with at least 20 hours duration. Error bars in the plot below correspond to the standard error of the mean. The C backend of our detector is by far the fastest implementation among all tested packages (note that the *y*-axis is logarithmically scaled). Runtime evaluation was performed on an [Intel® Xeon® Prozessor E5-2440 v2](https://ark.intel.com/content/www/us/en/ark/products/75263/intel-xeon-processor-e5-2440-v2-20m-cache-1-90-ghz.html) with 32 GiB RAM. No parallelization was used.
```

examples/benchmark/README.md
```diff
@@ -0,0 +1,39 @@
+# Heartbeat Detection Benchmarks
+This example allows to reproduce the benchmarks shown in the repo's main [`README.md`](https://github.com/cbrnr/sleepecg#readme).
```
cbrnr, line 2:
```suggestion
This example reproduces the benchmarks shown in the main [`README.md`](https://github.com/cbrnr/sleepecg#readme).
```

examples/benchmark/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Usage
+To run the benchmark, (create a virtual enviroment), install the requirements using
```
cbrnr, line 5:
```suggestion
To run the benchmark, create a virtual enviroment and install the requirements using
```

examples/benchmark/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+pip install -r requirements-benchmark.txt
+```
+
+and execute
+```
+python benchmark_detectors.py <benchmark>
+```
+where `<benchmark>` is a toplevel-key in `config.yml` (see section [Configuration](#configuration)). Possible options: `runtime`, `metrics`, `rri_similarity`.
```
cbrnr, line 14:
```suggestion
where `<benchmark>` is a toplevel-key in `config.yml` (see section [Configuration](#configuration)). The possible options are `runtime`, `metrics`, and `rri_similarity`.
```

examples/benchmark/README.md
```diff
@@ -0,0 +1,39 @@
...
+|detectors|`list[str]`||Detectors to be evaluated. For possible options, see [`utils._detector_dispatch`](https://github.com/cbrnr/sleepecg/blob/main/examples/benchmark/benchmark_detectors.py).|
+|signal_lengths|`list[int]`||Length in minutes to which each ECG signal should be sliced. If a signal is too short, it is skipped.|
+|max_distance|`float`|`0.1`|Maximum temporal distance in seconds between detected and annotated beats to count as a successful detection.|
+|suppress_warnings|`bool`|`False`|Whether to suppress warnings during detector execution.|
+|timeout|`int`|`600`|Number of seconds after which to attempt cancelling execution using SIGALRM (only on Unix).|
+|max_timeouts|`int`|`3`|Number of timeouts after which a detector is skipped completely.|
+|calc_rri_similarity|`bool`|`False`|Whether to calculate similarity measures between detected and annotated RR-intervals (computationally expensive for long signals).|
+
+## Known Issues:
+- Cancelling the `heartpy` detector for ltdb:15814:ECG2 doesn't work for `signal_lengths` larger than some value between 600 and 900 minutes. Program execution has to be terminated manually.
```
cbrnr, line 39:
```suggestion
- Cancelling the `heartpy` detector for ltdb:15814:ECG2 does not work for `signal_lengths` larger than some value between 600 and 900 minutes. Program execution has to be terminated manually.
```
What does this mean? Should users limit the length to 600 minutes for this data set then? I think there should be an explicit recommendation on how to avoid the situation that the program does not terminate.

examples/benchmark/config.yml
```diff
@@ -0,0 +1,61 @@
...
+    - mne
+    - wfdb_xqrs
+    - pyecg_pan_tompkins
+    - biosppy_hamilton
+    - heartpy
+    - neurokit2_nk
+    - neurokit2_kalidas2017
+    - sleepecg-c
+    - sleepecg-numba
+    - sleepecg-python
```
cbrnr, line 15: Maybe we can use dashes in all detectors (instead of underscores) to be consistent?

cbrnr (commented):
I did a very quick first pass mainly to read the text. I will take a closer look at the code in the next round (but in the meantime feel free to apply my changes if you agree).

examples/benchmark/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+and execute
+```
+python benchmark_detectors.py <benchmark>
+```
+where `<benchmark>` is a toplevel-key in `config.yml` (see section [Configuration](#configuration)). Possible options: `runtime`, `metrics`, `rri_similarity`.
+
+Plots can be created by executing
+```
+python plot_benchmark_results.py <results.csv>
```
cbrnr, line 18:
```suggestion
python plot_benchmark_results.py results.csv
```

examples/benchmark/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+which will save the plot to `results.svg` at the same location as `results.csv`. Plot types and labels for the provided benchmarks are selected based on the filename, so renaming may lead to errors.
+
+
+## Configuration
+A benchmark's configuration is specified below a unique top-level key in `config.yml`. Parameters:
```
cbrnr, line 24:
```suggestion
A benchmark configuration is specified below a unique top-level key in `config.yml`. Parameters:
```

examples/benchmark/README.md
```diff
@@ -0,0 +1,39 @@
...
+|Key|Type|Default|Description|
```
cbrnr, line 25: Should the values in the *Key* column be rendered as code (backticks)?

examples/benchmark/README.md
```diff
@@ -0,0 +1,39 @@
...
+|db_slug|`str`||Which dataset to use for the evaluation. Possible values: [`mitdb`](https://physionet.org/content/mitdb/1.0.0/), [`ltdb`](https://physionet.org/content/mitdb/1.0.0/), [`gudb`](https://physionet.org/content/mitdb/1.0.0/).|
+|detectors|`list[str]`||Detectors to be evaluated. For possible options, see [`utils._detector_dispatch`](https://github.com/cbrnr/sleepecg/blob/main/examples/benchmark/benchmark_detectors.py).|
+|signal_lengths|`list[int]`||Length in minutes to which each ECG signal should be sliced. If a signal is too short, it is skipped.|
+|max_distance|`float`|`0.1`|Maximum temporal distance in seconds between detected and annotated beats to count as a successful detection.|
+|suppress_warnings|`bool`|`False`|Whether to suppress warnings during detector execution.|
+|timeout|`int`|`600`|Number of seconds after which to attempt cancelling execution using SIGALRM (only on Unix).|
+|max_timeouts|`int`|`3`|Number of timeouts after which a detector is skipped completely.|
+|calc_rri_similarity|`bool`|`False`|Whether to calculate similarity measures between detected and annotated RR-intervals (computationally expensive for long signals).|
```
cbrnr, line 36:
```suggestion
|calc_rri_similarity|`bool`|`False`|Whether to calculate similarity measures between detected and annotated RR intervals (computationally expensive for long signals).|
```

examples/benchmark/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Known Issues:
```
cbrnr, line 38:
```suggestion
## Known Issues
```

examples/benchmark/utils.py
```diff
@@ -0,0 +1,206 @@
...
+        clean_ecg = neurokit2.ecg.ecg_clean(ecg, int(fs), method='kalidas2017')
+        detection = neurokit2.ecg.ecg_findpeaks(clean_ecg, int(fs), method='kalidas2017')['ECG_R_Peaks']  # noqa
+    elif detector == 'sleepecg-c':
+        detection = sleepecg.detect_heartbeats(ecg, fs, backend='c')
+    elif detector == 'sleepecg-numba':
+        detection = sleepecg.detect_heartbeats(ecg, fs, backend='numba')
+    elif detector == 'sleepecg-python':
+        detection = sleepecg.detect_heartbeats(ecg, fs, backend='python')
+    else:
+        raise ValueError(f'Unknown QRS-detector: {detector}')
```
cbrnr, line 106:
```suggestion
        raise ValueError(f'Unknown QRS detector: {detector}')
```

examples/benchmark/utils.py
```diff
@@ -0,0 +1,206 @@
...
+    max_distance: float,
+    calc_rri_similarity: bool,
+    timeout: int,
+    max_timeouts: int,
+) -> dict[str, Any]:
+    """
+    Evaluate a heartbeat detector on a given annotated ECG record.
+
+    Optionally, similarity measures between detected and annotated
+    RR-intervals can be calculated. As this requires interpolation, it may
```
cbrnr, line 123:
```suggestion
    RR intervals can be calculated. As this requires interpolation, it may
```

examples/benchmark/utils.py
```diff
@@ -0,0 +1,206 @@
...
+        'num_samples': len(ecg),
+        'detector': detector,
+        'max_distance': max_distance,
+        'runtime': runtime,
+        'TP': len(TP),
+        'FP': len(FP),
+        'FN': len(FN),
+        'pearsonr': rri_similarity.pearsonr if calc_rri_similarity else np.nan,
+        'spearmanr': rri_similarity.spearmanr if calc_rri_similarity else np.nan,
+        'rmse': rri_similarity.rmse if calc_rri_similarity else np.nan,
```
cbrnr, line 204: I wonder if we should only include those entries `if calc_rri_similarity` is `True`. Or is it important to always have exactly the same set of dict keys?

examples/benchmark/README.md
hofaflo: I just tried to find out why that happens - with no success, but apparently it isn't necessarily only above 600 minutes. So maybe like this:

```suggestion
- Cancelling the `heartpy` detector for ltdb:15814:ECG2 does not work for long `signal_lengths` (possibly starting at some value between 600 and 900 minutes). To avoid having to kill the process manually, we recommend leaving `heartpy` out of the benchmark or evaluating signal lengths up to 600 minutes only.
```

cbrnr:
I get an error when running the `runtime` benchmark (possible also for the other ones):

```
$ python benchmark_detectors.py runtime                   
Downloading ltdb: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 7/7 [01:15<00:00, 10.72s/it]
Storing results to /Users/clemens/Repositories/sleepecg/examples/benchmark/results/runtime__2021_08_18__11_55_55.csv
Traceback (most recent call last):
  File "/Users/clemens/Repositories/sleepecg/examples/benchmark/benchmark_detectors.py", line 41, in <module>
    with open(csv_filepath, 'w', newline='') as csv_file:
FileNotFoundError: [Errno 2] No such file or directory: './results/runtime__2021_08_18__11_55_55.csv'
```

Maybe this is because the `results` folder does not exist?

hofaflo:
> I get an error when running the `runtime` benchmark (possible also for the other ones):
> 
> ```
> $ python benchmark_detectors.py runtime                   
> Downloading ltdb: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 7/7 [01:15<00:00, 10.72s/it]
> Storing results to /Users/clemens/Repositories/sleepecg/examples/benchmark/results/runtime__2021_08_18__11_55_55.csv
> Traceback (most recent call last):
>   File "/Users/clemens/Repositories/sleepecg/examples/benchmark/benchmark_detectors.py", line 41, in <module>
>     with open(csv_filepath, 'w', newline='') as csv_file:
> FileNotFoundError: [Errno 2] No such file or directory: './results/runtime__2021_08_18__11_55_55.csv'
> ```
> 
> Maybe this is because the `results` folder does not exist?

right, fixed

cbrnr:
I'm a big fan of `pathlib`, what about using that instead of `os`?

cbrnr:
Also, Pylance is showing a warning (Import "utils" could not be resolved) for the import

```
from utils import evaluate_single, reader_dispatch
```

Should we ignore it? The warning disappears when you make that a relative import, but I don't know if it is common to use relative imports when you are not referring to a package.

cbrnr:
The first call to our Numba detector is very slow because of the JIT compilation, so the first benchmark result (5 min signal length) is not representative. We should trigger the compilation manually as a first step.

hofaflo:
> Also, Pylance is showing a warning (Import "utils" could not be resolved) for the import
> 
> ```
> from utils import evaluate_single, reader_dispatch
> ```
> 
> Should we ignore it? The warning disappears when you make that a relative import, but I don't know if it is common to use relative imports when you are not referring to a package.

The warning disappears but you'll get this instead ;)
```
ImportError: attempted relative import with no known parent package
```

cbrnr:
I guess we'll leave the import as is and just ignore the warning then. Pylance is still weird sometimes.

cbrnr:
For signal length 10 minutes I get this warning:

```
  mean runtime:  0.40434s ->   25 minutes analyzed per second # biosppy-hamilton
/Users/clemens/Repositories/sleepecg/.venv/lib/python3.9/site-packages/scipy/interpolate/fitpack2.py:280: UserWarning: 
The maximal number of iterations maxit (set to 20 by the program)
allowed for finding a smoothing spline with fp=s has been reached: s
too small.
There is an approximation returned but the corresponding weighted sum
of squared residuals does not satisfy the condition abs(fp-s)/s < tol.
  warnings.warn(message)
```

Is this expected/OK?

hofaflo:
> For signal length 10 minutes I get this warning:
> 
> ```
>   mean runtime:  0.40434s ->   25 minutes analyzed per second # biosppy-hamilton
> /Users/clemens/Repositories/sleepecg/.venv/lib/python3.9/site-packages/scipy/interpolate/fitpack2.py:280: UserWarning: 
> The maximal number of iterations maxit (set to 20 by the program)
> allowed for finding a smoothing spline with fp=s has been reached: s
> too small.
> There is an approximation returned but the corresponding weighted sum
> of squared residuals does not satisfy the condition abs(fp-s)/s < tol.
>   warnings.warn(message)
> ```
> 
> Is this expected/OK?

I'd say yes, `heartpy` manages to detect something nonetheless 🤷.

hofaflo:
> The first call to our Numba detector is very slow because of the JIT compilation, so the first benchmark result (5 min signal length) is not representative. We should trigger the compilation manually as a first step.

where would you suggest to do that?

examples/benchmark/README.md
hofaflo: I'd keep the `<>` since it's a placeholder for e.g. `rri_similarity__2021_08_18__12_23_07.csv`.

cbrnr:
> where would you suggest to do that?

Maybe run it on a very short data segment of a data set that is part of the benchmark (and of course only do that if the user included `sleepecg-numba`)?

examples/benchmark/README.md
cbrnr: This sounds good. But does that mean that `heartpy` does not *work* for long signals or that computation just cannot be *interrupted*?

hofaflo: It can't be interrupted by SIGALRM at least... I guess it does work, but would likely take hours for that specific signal.

examples/benchmark/README.md
cbrnr: Ah, got it. I thought it was the literal name.

examples/benchmark/README.md
cbrnr: So how does this influence the benchmark script? Does it take ages or is this detector interrupted? Or is it not run for durations > 600 minutes?

examples/benchmark/requirements-benchmark.txt
```diff
@@ -0,0 +1,14 @@
+# Evaluation
+numpy>=1.20.0
+pandas>=1.3.1
+plotly>=5.1.0
+pyyaml>=5.4.1
+sleepecg>=0.2.0
+
```
cbrnr, line 7: Exporting figures requires `kaleido`, so this package should be added here (not sure about the min version).

cbrnr:
I feel like we could add a simpler visualization for the runtime benchmark. It looks like algorithms are pretty independent of signal length, so we could add a bar chart showing algorithm runtime per minute of signal. Or maybe analyzed minutes per second (I think I'd prefer the former though). WDYT?

examples/benchmark/README.md
hofaflo: I've never waited long enough to find out if it actually finishes, but terminated the script manually.

hofaflo:
> I feel like we could add a simpler visualization for the runtime benchmark. It looks like algorithms are pretty independent of signal length, so we could add a bar chart showing algorithm runtime per minute of signal. Or maybe analyzed minutes per second (I think I'd prefer the former though). WDYT?

Most of them are independent, just `heartpy` and `biosspy-hamilton` get slower for longer signals. Therefore I'd prefer to keep the current visualization.
runtime__perminute

cbrnr:
OK! Can we add that figure you just showed?

Final points:
- How does sampling frequency influence the runtime? Hz is missing in the figure caption in `fs=128`.
- I don't like that the benchmark needs to be terminated manually because of two slow detectors. Can we somehow make it so that users can just run the benchmark and it completes after some finite time?

hofaflo:
> OK! Can we add that figure you just showed?

Done, do you want to show it in the readme as well?

> * How does sampling frequency influence the runtime? Hz is missing in the figure caption in `fs=128`.

It's just included for the sake of completeness (although `wfdb-xqrs` does get slower with increasing fs, but that's not shown here).

> * I don't like that the benchmark needs to be terminated manually because of two slow detectors. Can we somehow make it so that users can just run the benchmark and it completes after some finite time? Maybe just limit the maximum time to 600 minutes in `config.yml`?

The first option would likely require starting a second process watching over the main one. I'd go with your second idea, since the plots will still show the performance decreasing with signal length.

> * I think some information on how many data sets are used from the used database would be helpful. In the code, you seem to limit this to 25, although for the runtime benchmark there are only 7 * 2 data sets available anyway.

True, that shouldn't be there 😅. The benchmark script will now print how many records have been read.

cbrnr:
No need to include it in the main `README.md`. But we could drop the last code block (detector calls) and instead mention at the beginning of the Performance Evaluation section that the code is available in `examples/benchmark`.

cbrnr:
> True, that shouldn't be there 😅. The benchmark script will now print how many records have been read.

I'm still confused, because now it prints "15 records". There are 7 files, each of which contains 2 ECG channels, and one possibly 3. Could the message be more informative and include these details? At least the number of files should be printed (i.e. "7 files containing 15 records").

README.md
```diff
@@ -29,7 +29,9 @@ detection = detect_heartbeats(ecg, fs)
 
 
 ### Performance evaluation
-We evaluated detector runtime using slices of different lengths from [LTDB](https://physionet.org/content/ltdb/1.0.0/) records with at least 20 hours duration. Error bars in the plot below correspond to the standard error of the mean. Our detector (we only show the C backend) is by far the fastest implementation among all tested packages (note that the *y*-axis is logarithmically scaled).
+All code used for performance evaluation is available in [`examples/benchmark`](https://github.com/cbrnr/sleepecg/tree/main/examples/benchmark). The used package versions are listed in [`requirements-benchmark.txt`](https://github.com/cbrnr/sleepecg/blob/main/examples/benchmark/requirements-benchmark.py)
```
cbrnr, line 32:
```suggestion
All code used for performance evaluation is available in [`examples/benchmark`](https://github.com/cbrnr/sleepecg/tree/main/examples/benchmark). The used package versions are listed in [`requirements-benchmark.txt`](https://github.com/cbrnr/sleepecg/blob/main/examples/benchmark/requirements-benchmark.py).
```

hofaflo:
> > True, that shouldn't be there 😅. The benchmark script will now print how many records have been read.
> 
> I'm still confused, because now it prints "15 records". There are 7 files, each of which contains 2 ECG channels, and one possibly 3. Could the message be more informative and include these details? At least the number of files should be printed (i.e. "7 files containing 15 records").

It could, though that would require changes in `io.ecg_readers`, since those are generators yielding from two nested loops (over the files and over the signals included in one file).

cbrnr:
> It could, though that would require changes in io.ecg_readers, since those are generators yielding from two nested loops (over the files and over the signals included in one file).

OK, then let's leave it as is (this is maybe something for later, I really think this would be valuable information). But I'm still curious: how many ECG signals does each of the seven files have? Why don't all of them just have two?

hofaflo:
> OK, then let's leave it as is (this is maybe something for later, I really think this would be valuable information). 

👍 

> But I'm still curious: how many ECG signals does each of the seven files have? Why don't all of them just have two?
 
Most have two, but 15814 has three for some reason... there isn't any information on [PhysioNet](https://physionet.org/content/ltdb/1.0.0/) :/

cbrnr (commented):
Now finished reviewing `utils.py`, the main point is that we should probably remove the timeout stuff.

examples/benchmark/utils.py
```diff
@@ -0,0 +1,215 @@
...
+        beats to count as a successful detection.
+    calc_rri_similarity : bool
+        If `True`, calculate similarity measures between detected and
+        annotated RR-intervals (computationally expensive for long
+        signals).
+    timeout : int
+        Number of seconds after which to attempt cancelling execution using
+        SIGALRM.
+    max_timeouts : int
+        Number of timeouts after which a detector is skipped completely.
```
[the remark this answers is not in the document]
hofaflo: I had messed the defaults in the readme, those are fixed now.

If a detector times out, it carries on with the next record, so it won't run forever.

`signal.alarm(timeout)` asks the OS for a `SIGALRM` after `timeout seconds`, which is then handled by `sigalrm_handler`, which in turn raises the `TimeoutError`. The triggering time limit is set in `config.yml` using the key `timeout`. 

We'd need to limit it to 300 minutes then (see "Known Issues" at the bottom of `README.md`), since `wfdb-xqrs` behaves strangely for some records. Or figure out the maximum signal length where it still works.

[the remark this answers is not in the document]
cbrnr: > I had messed the defaults in the readme, those are fixed now.

👍 

> If a detector times out, it carries on with the next record, so it won't run forever.

This isn't entirely clear from reading the docs, but it now makes sense. The relationship between `timeout` and `max_timeouts` isn't immediately clear. Also, `timeout=0` is a special case which needs to be documented.

> `signal.alarm(timeout)` asks the OS for a `SIGALRM` after `timeout` seconds, which is then handled by `sigalrm_handler`, which in turn raises the `TimeoutError`. The triggering time limit is set in `config.yml` using the key `timeout`.

Got it. But all runs in the config file use the default of zero.

> We'd need to limit it to 300 minutes then (see "Known Issues" at the bottom of `README.md`), since `wfdb-xqrs` behaves strangely for some records. Or figure out the maximum signal length where it still works.

I'm tempted to do that. The whole timeout thing is just because this detector cannot handle long signals. Why don't we hard-code the max signal length for `wfdb-xqrs`? Then we could still use lengths up to 600 minutes for all other detectors.

Alternative suggestion: we could limit `timeout` and `max_timeouts` to values that would lead to skipping `wfdb-xqrs` for 600+ minutes, but not others.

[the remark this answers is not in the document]
cbrnr: But given that canceling with `SIGALRM` doesn't work for this detector kind of means that this isn't really that useful.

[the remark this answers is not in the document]
hofaflo: > Got it. But all runs in the config file use the default of zero.

Right, that is because apparently I was a bit too enthusiastic when cleaning up that file 🙈 

Maybe a summary of the issues will help here:
- `heartpy ` 
  - has extreme variability in it's runtime (e.g. for 600 signal minutes, runtimes are between 5 and 1135 seconds (not a typo))
  - above a signal length somewhere between 600 and 900 minutes, ltdb:15184:ECG2 takes at least several hours
  - can't be interrupted by `SIGALRM` (or at least the part taking so long can't be)
- `wfdb-xqrs`
  - mainly has low runtime variability
  - above a signal length somewhere between 300 and 600 minutes, ltdb:14134:ECG2 and ltdb:14184:ECG2 take much longer than the other records at the same length
  - can be interrupted by `SIGALRM`
- `biosppy-hamilton`
  - also has low runtime variability
  - doesn't scale well with signal length
  - can be interrupted by `SIGALRM`

So the simplest solution would likely be to use only signal lengths up to 300 minutes and add these notes to the readme for anyone who would want to try longer signals. This would result in the following plots:
runtime__300
runtime_per_minute__300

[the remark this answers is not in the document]
cbrnr: I think this is a good solution. We could even create another entry in the config (e.g. `runtime-long`), where we exclude these slow detectors and explain why. Then we can have another figure showing longer signal lengths only, which is particularly interesting because we're doing sleep ECG analysis, so those algorithms cannot be used for this purpose.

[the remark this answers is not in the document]
cbrnr: But I think it will make the benchmark script much simpler if we get rid of the timeout and canceling business.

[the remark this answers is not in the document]
hofaflo: > I think this is a good solution. We could even create another entry in the config (e.g. `runtime-long`), where we exclude these slow detectors and explain why. Then we can have another figure showing longer signal lengths only, which is particularly interesting because we're doing sleep ECG analysis, so those algorithms cannot be used for this purpose.

Since all detectors which would be included in a `runtime-long` benchmark scale more or less linearly with signal length, I think we could leave it at the 300 minute max. benchmark.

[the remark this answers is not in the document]
cbrnr: OK, then let's make sure to mention which detectors cannot be used for signals longer than about five hours in the docs!

examples/benchmark/plot_benchmark_results.py
```diff
@@ -0,0 +1,99 @@
...
+            'signal_len': 'signal length in hours',
+            'mean_runtime_per_minute': 'mean runtime per minute in s',
+        },
+        title=f'Mean detector runtime per signal minute for {db_slug.upper()} (fs=128Hz)',
+        width=800,
+        height=600,
+        render_mode='svg',
+    )
+    fig.update_yaxes(rangemode='tozero')
+    fig.write_image(plot_filepath.with_name(plot_filepath.name.replace('runtime', 'runtime_per_minute')))  # noqa
```
cbrnr, line 70: For some reason, the first figure is correctly called `runtime__ltdb__2021_08_21__19_49_33.csv`, but the second figure is missing the `db_slug` part because it is called `runtime_per_minute__2021_08_20__10_44_04.svg`.

Now that the first plot is computed much faster, I am now tending towards dropping the runtime per minute figure again. Sorry about that, but it makes this file simpler again.

cbrnr (changes requested):
I have a few minor comments, including my change of mind regarding the per minute figure (which I would like to remove again).

Also, I get an error when trying to plot the results I successfully calculated:

```
$ python plot_benchmark_results.py results/runtime__ltdb__2021_08_21__19_49_33.csv
Traceback (most recent call last):
  File "/Users/clemens/Repositories/sleepecg/.venv/lib/python3.9/site-packages/pandas/core/indexes/base.py", line 3361, in get_loc
    return self._engine.get_loc(casted_key)
  File "pandas/_libs/index.pyx", line 76, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/index.pyx", line 108, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 5198, in pandas._libs.hashtable.PyObjectHashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 5206, in pandas._libs.hashtable.PyObjectHashTable.get_item
KeyError: 'fs'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/clemens/Repositories/sleepecg/examples/benchmark/plot_benchmark_results.py", line 32, in <module>
    fs = results['fs'][0]
  File "/Users/clemens/Repositories/sleepecg/.venv/lib/python3.9/site-packages/pandas/core/frame.py", line 3455, in __getitem__
    indexer = self.columns.get_loc(key)
  File "/Users/clemens/Repositories/sleepecg/.venv/lib/python3.9/site-packages/pandas/core/indexes/base.py", line 3363, in get_loc
    raise KeyError(key) from err
KeyError: 'fs'
```

The column is present in the file.

After we've resolved all of these issues, I will take a final look and comment on the figures.

examples/benchmark/utils.py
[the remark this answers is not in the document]
hofaflo: The "strange behaviours" are mentioned in "Known Issues" in the benchmark readme and I think what runtime to expect from which detector should be clear from the plot. Would you like to add anything?

examples/benchmark/plot_benchmark_results.py
```diff
@@ -0,0 +1,99 @@
...
+    fig = px.box(
+        results,
+        y='pearsonr',
+        color='detector',
+        title=f'Pearson correlation coefficient for RRI timeseries from {db_slug.upper()}',
+    )
+    fig.write_image(plot_filepath, scale=1.5)
+
+else:
+    raise ValueError(f'No plotting strategy defined for {results_filepath}')
```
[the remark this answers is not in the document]
hofaflo: Did you mean that?
```suggestion
    raise ValueError(f'No plotting strategy defined for {results_filepath}.')
```

examples/benchmark/utils.py
```diff
@@ -0,0 +1,179 @@
...
+    -------
+    dict[str, Any]
+        A dictionary containing evaluation results.
+    """
+    signal_len_samples = int(signal_len * record.fs*60)
+    ecg = record.ecg[:signal_len_samples]
+    annotation = record.annotation[record.annotation < signal_len_samples]
+    fs = int(record.fs)
+
+    try:
```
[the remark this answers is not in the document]
hofaflo: `heartpy` _raises_ a `BadSignalWarning` for mitdb:105:V1 🤷

examples/benchmark/plot_benchmark_results.py
```diff
@@ -0,0 +1,99 @@
...
+import plotly.express as px
+
+results_filepath = Path(sys.argv[1])
+benchmark, db_slug, *_ = results_filepath.stem.split('__')
+plot_filepath = results_filepath.with_suffix('.svg')
+results = pd.read_csv(results_filepath).sort_values('detector')
+
+if benchmark == 'runtime':
+    results['signal_len'] = results['num_samples'] / (results['fs'] * 3600)
+    results['runtime_per_minute'] = results['runtime'] / results['signal_len'] / 60  # noqa
```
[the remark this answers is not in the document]
hofaflo: Irrelevant now, but for completeness' sake: because I had the calculation wrong at first and that line was longer^^

examples/benchmark/plot_benchmark_results.py
hofaflo: I'm not sure what you mean by "the first plot being computed faster"? But agreed, it's gone.

cbrnr: Never mind, I don't know what I meant either 😄 – the important thing is that the second figure is now gone again!

examples/benchmark/utils.py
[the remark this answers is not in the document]
cbrnr: Is that the signal that also leads to the warning in the benchmark? What does this `BadSignalWarning` mean? If it is just for this signal, and we know that the warning can be ignored, let's just catch this one specifically and keep the rest of our code simple (i.e. no error message storing etc.).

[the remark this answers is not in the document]
cbrnr: Ah, and it is just a warning and not an exception, so no need for the `try/except` block (which is gone in the latest commit anyway).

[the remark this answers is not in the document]
hofaflo: > And it is just a warning and not an exception, so no need for the `try/except` block.

It is derived from `UserWarning`, which is a subclass of `Exception`, which allows it to actually be _raised_ [here](https://github.com/paulvangentcom/heartrate_analysis_python/blob/42534d74d31ec6575a5cc0111787c21339e6002b/heartpy/peakdetection.py#L305-L309).

> Is that the signal that also leads to the warning in the benchmark? 

The one you posted [here](https://github.com/cbrnr/sleepecg/pull/22#issuecomment-901000724) was caused by an ltdb record (for which there's not `BadSignalWarning`). That warning also appears for at least one mitdb record, but not for the one which leads to the `BadSignalWarning`.

> What does this `BadSignalWarning` mean? 

The error message is not too specific on that:
```
----------------
Could not determine best fit for given signal. Please check the source signal.
 Probable causes:
- detected heart rate falls outside of bpmmin<->bpmmax constraints
- no detectable heart rate present in signal
- very noisy signal (consider filtering and scaling)
If you're sure the signal contains heartrate data, consider filtering and/or scaling first.
----------------
```

The signal isn't noisy, it just has large negative peaks where the QRS-complex should be, so maybe that's the problem?
image


> If it is just for this signal, and we know that the warning can be ignored, let's just catch this one specifically and keep the rest of our code simple (i.e. no error message storing etc.).

It's the only signal in ltdb, mitdb and gudb which causes this exception - catching it specifically is a good point, yes!

examples/benchmark/utils.py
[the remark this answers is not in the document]
cbrnr: Maybe mention that `heartpy` fails on one signal which we're skipping by catching the `BadSignalWarning`?

cbrnr:
Final two comments:
- Let's mention the record where `heartpy` fails with the `BadSignalWarning` explicitly in the "Known issues" section.
- I'd scale the y-axis of the correlation figure from -1 to +1.

Then this if really ready to be merged! Thanks for your patience @hofaflo!

hofaflo:
Done, thanks for your thoroughness! :)

examples/benchmark/README.md
```diff
@@ -0,0 +1,39 @@
+# Heartbeat Detection Benchmarks
```
cbrnr, line 1:
```suggestion
# Heartbeat detection benchmarks
```

examples/benchmark/README.md
```diff
@@ -0,0 +1,39 @@
...
+|`data_dir`|`str`||Path where all datasets are stored.|
+|`outfile_dir`|`str`||Path where the evaluation results should be stored.|
+|`db_slug`|`str`||Which dataset to use for evaluation. Possible values: [`mitdb`](https://physionet.org/content/mitdb/1.0.0/), [`ltdb`](https://physionet.org/content/ltdb/1.0.0/), [`gudb`](https://github.com/berndporr/ECG-GUDB).|
+|`detectors`|`list[str]`||Detectors to be evaluated. For possible options, see [`utils._detector_dispatch`](https://github.com/cbrnr/sleepecg/blob/main/examples/benchmark/benchmark_detectors.py).|
+|`signal_lengths`|`list[int]`||Length in minutes to which each ECG signal should be sliced. If a signal is too short, it is skipped.|
+|`max_distance`|`float`|`0.1`|Maximum temporal distance in seconds between detected and annotated beats to count as a successful detection.|
+|`suppress_warnings`|`bool`|`False`|Whether to suppress warnings during detector execution.|
+|`calc_rri_similarity`|`bool`|`False`|Whether to calculate similarity measures between detected and annotated RR intervals (computationally expensive for long signals).|
+
+## Known Issues
```
cbrnr, line 36:
```suggestion
## Known issues
```

cbrnr (approved):

cbrnr:
Before I merge this (just waiting for the CIs to go green), I have two questions:

1. I was wondering if people will find the benchmark folder in their locally installed package. If someone wants to run the benchmark, what is the easiest way to get to the folder?
2. Currently, the downloaded data sets go into the `examples/benchmark` folder. Is this a good location? What if people download some database files elsewhere? It's unlikely that they will put it in the `examples/datasets` folder. Should we create a default location somewhere in the home directory (e.g. `~/.sleepecg`)?

We also need to add the `examples` (and maybe also `doc` and possibly others) to the sdist I think.

hofaflo:
> 1. I was wondering if people will find the benchmark folder in their locally installed package. If someone wants to run the benchmark, what is the easiest way to get to the folder?

I guess cloning the repo would be the easiest way right now 🤔 

> 2. Currently, the downloaded data sets go into the `examples/benchmark` folder. Is this a good location? What if people download some database files elsewhere? It's unlikely that they will put it in the `examples/datasets` folder. Should we create a default location somewhere in the home directory (e.g. `~/.sleepecg`)?

I've chosen that location since it will exist for everyone and not mess up any other folder structure they might have created. Changing it in `config.yml` to find preexisting datasets should be straightforward, but we can also set `~/.sleepecg` as the default, if you prefer!

> We also need to add the `examples` (and maybe also `doc` and possibly others) to the sdist I think.

Yep, your idea from [here](https://github.com/cbrnr/sleepecg/issues/6#issuecomment-902633389) should take care of that, right?

cbrnr:
> I guess cloning the repo would be the easiest way right now 🤔

True. Should we mention this explicitly somewhere?

> I've chosen that location since it will exist for everyone and not mess up any other folder structure they might have created. Changing it in `config.yml` to find preexisting datasets should be straightforward, but we can also set `~/.sleepecg` as the default, if you prefer!

But what folder structure would creating and using `~/.sleepecg` mess up? If people continue to use SleepECG after running the benchmark, they would immediately have all the already downloaded data sets if the default location was `~/.sleepecg`.

cbrnr:
> Yep, your idea from here should take care of that, right?

Yes, but I'd add each required directory separately.

hofaflo:
> True. Should we mention this explicitly somewhere?

Not sure, what's the usual way of using package examples?

> But what folder structure would creating and using `~/.sleepecg` mess up?

None, but I only though about e.g. `~/datasets` :D

> If people continue to use SleepECG after running the benchmark, they would immediately have all the already downloaded data sets if the default location was `~/.sleepecg`.

👍

hofaflo:
Would you save the benchmark results to `~/.sleepecg/benchmark_results/` then?

cbrnr:
> Not sure, what's the usual way of using package examples?

Not sure either, but maybe we can provide a console entry point in `setup.cfg` which runs the benchmarks? Then users don't have to know where the benchmark code is located on their computer. For example, we could call it `sleepecg_benchmark`?

> None, but I only though about e.g. ~/datasets :D

OK, let's make `~/.sleepecg` the default then. We could even add that as a default argument for the readers?

> Would you save the benchmark results to ~/.sleepecg/benchmark_results/ then?

Hm, tough questions. Having data in that hidden folder is fine because most users will never want to access the data directly. However, users will want to open their result CSV and SVG files, and that hidden location is not ideal. Maybe just store them in the CWD?

hofaflo:
> Not sure either, but maybe we can provide a console entry point in `setup.cfg` which runs the benchmarks? Then users don't have to know where the benchmark code is located on their computer. For example, we could call it `sleepecg_benchmark`?

I like the idea, though that would require `examples/benchmark` to be a package so imports will work and I'm not sure what that would do to the package installation process... I'd stick with the "clone and call the script" approach.

> OK, let's make `~/.sleepecg` the default then. 

I opted for `~/.sleepecg/datasets` now, in case we want to put things other than datasets there as well.

> We could even add that as a default argument for the readers?

Good idea, it's on my TODO-list.

> Maybe just store them in the CWD?

👍

examples/benchmark/benchmark_detectors.py
```diff
@@ -19,7 +19,7 @@
     print('Usage: python benchmark_detectors.py <benchmark>')
     exit()
 
-with open('config.yml') as config_file:
+with open(Path(__file__).parent / 'config.yml') as config_file:
```
cbrnr, line 22: Why is this necessary?

examples/README.md
```diff
@@ -0,0 +1,12 @@
+# Examples
+
+To run the provided examples, download or clone the [GitHub Repository](https://github.com/cbrnr/sleepecg) and execute the scripts in this directory or its subdirectories. E.g.
```
cbrnr, line 3:
```suggestion
To run the provided examples, download or clone the [GitHub Repository](https://github.com/cbrnr/sleepecg) and execute the scripts in this directory or its subdirectories.
```

examples/README.md
```diff
@@ -0,0 +1,12 @@
...
+- Heartbeat detection demo:
+    ```
+    ~/sleepecg/examples> python heartbeat_detection.py
```
cbrnr, line 6:
```suggestion
    sleepecg/examples> python heartbeat_detection.py
```

examples/README.md
```diff
@@ -0,0 +1,12 @@
...
+    ```
+
+- Benchmark heartbeat detector runtime (more info [here](https://github.com/cbrnr/sleepecg/tree/main/examples/benchmark))
```
cbrnr, line 9:
```suggestion
- Benchmark heartbeat detector runtime (more info [here](https://github.com/cbrnr/sleepecg/tree/main/examples/benchmark)):
```

examples/README.md
```diff
@@ -0,0 +1,12 @@
...
+    ```
+    ~/sleepecg/examples/benchmark> python benchmark_detectors.py runtime
```
cbrnr, line 11:
```suggestion
    sleepecg/examples/benchmark> python benchmark_detectors.py runtime
```

examples/benchmark/benchmark_detectors.py
hofaflo: in case someone calls the benchmark script from somewhere else than the `examples/benchmark` folder... feel free to revert if you think that's not necessary :)

cbrnr: I don't think this is necessary, because in our instructions we explicitly mention to change into the directory first.

cbrnr:
Thanks @hofaflo! Great work in getting this PR in 🚀!

PR #26, opened 2021-08-20
Rename test folder

cbrnr:
Tests are automatically included in the sdist if the folder name is "test" (not "tests") (see [here](https://packaging.python.org/guides/using-manifest-in/#how-files-are-included-in-an-sdist)). @hofaflo I don't think this has any negative side effects, because pytest doesn't put any constraints on folder names (it runs `test_*.py` and `*_test.py` files in any folder).


hofaflo:
Nice solution, thanks @cbrnr!

PR #28, opened 2021-08-20
Add `io.read_mesa`

hofaflo:
Add a reader function for the [MESA](https://sleepdata.org/datasets/mesa) dataset. 

This PR also establishes the data structures which should be used by all other readers:
- `SleepRecord`: A `dataclass` for storing a single record consisting of heartbeat times and annotates stages.
- `SleepStage`: An `IntEnum` mapping AASM sleep stage identifiers to integers (will be required for classification and plotting).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [UNRELEASED]
 ### Added
 - Configuration via a YAML file ([#39](https://github.com/cbrnr/sleepecg/pull/39) by [Florian Hofer](https://github.com/hofaflo))
+- Function for reading data from the [MESA](https://sleepdata.org/datasets/mesa) datasets ([#28](https://github.com/cbrnr/sleepecg/pull/28) by [Florian Hofer](https://github.com/hofaflo))
 
 ## [0.3.0] - 2021-08-25
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## [UNRELEASED]
 ### Added
 - Configuration via a YAML file ([#39](https://github.com/cbrnr/sleepecg/pull/39) by [Florian Hofer](https://github.com/hofaflo))
-- Function for reading data from the [MESA](https://sleepdata.org/datasets/mesa) datasets ([#28](https://github.com/cbrnr/sleepecg/pull/28) by [Florian Hofer](https://github.com/hofaflo))
+- Add function for reading data from the [MESA](https://sleepdata.org/datasets/mesa) datasets ([#28](https://github.com/cbrnr/sleepecg/pull/28) by [Florian Hofer](https://github.com/hofaflo))
 
 ## [0.3.0] - 2021-08-25
 ### Added
--- a/sleepecg/io/__init__.py
+++ b/sleepecg/io/__init__.py
@@ -2,9 +2,10 @@
 
 from .ecg_readers import read_gudb, read_ltdb, read_mitdb
 from .nsrr import download_nsrr, set_nsrr_token
-from .sleep_readers import read_mesa
+from .sleep_readers import SleepRecord, read_mesa
 
 __all__ = [
+    'SleepRecord',
     'read_gudb',
     'read_ltdb',
     'read_mitdb',
--- a/sleepecg/io/__init__.py
+++ b/sleepecg/io/__init__.py
@@ -2,11 +2,13 @@
 
 from .ecg_readers import read_gudb, read_ltdb, read_mitdb
 from .nsrr import download_nsrr, set_nsrr_token
+from .sleep_readers import read_mesa
 
 __all__ = [
     'read_gudb',
     'read_ltdb',
     'read_mitdb',
+    'read_mesa',
     'set_nsrr_token',
     'download_nsrr',
 ]
--- a/sleepecg/io/__init__.py
+++ b/sleepecg/io/__init__.py
@@ -2,10 +2,9 @@
 
 from .ecg_readers import read_gudb, read_ltdb, read_mitdb
 from .nsrr import download_nsrr, set_nsrr_token
-from .sleep_readers import SleepRecord, read_mesa
+from .sleep_readers import read_mesa
 
 __all__ = [
-    'SleepRecord',
     'read_gudb',
     'read_ltdb',
     'read_mitdb',
--- a/sleepecg/io/sleep_readers.py
+++ b/sleepecg/io/sleep_readers.py
@@ -19,8 +19,6 @@ from .nsrr import get_nsrr_url, list_nsrr
 from .utils import download_file
 
 __all__ = [
-    'SleepRecord',
-    'SleepStage',
     'read_mesa',
 ]
 
--- a/sleepecg/io/sleep_readers.py
+++ b/sleepecg/io/sleep_readers.py
@@ -185,7 +185,7 @@ def read_mesa(
     if not offline:
         download_url = get_nsrr_url(DB_SLUG)
 
-    db_dir = Path(data_dir) / DB_SLUG
+    db_dir = Path(data_dir).expanduser() / DB_SLUG
     annotations_dir = db_dir / ANNOTATION_DIRNAME
     edf_dir = db_dir / EDF_DIRNAME
     heartbeats_dir = db_dir / HEARTBEATS_DIRNAME
--- a/sleepecg/io/sleep_readers.py
+++ b/sleepecg/io/sleep_readers.py
@@ -146,7 +146,7 @@ def read_mesa(
     polysomnography data and an `.xml` file containing annotated events.
     Since the entire MESA dataset requires about 385 GB of disk space,
     `.edf` files can be deleted after heartbeat times have been extracted.
-    Heartbeat times are cached in a `.npy` file in
+    Heartbeat times are cached in an `.npy` file in
     `<data_dir>/mesa/preprocessed/heartbeats`.
 
     Parameters
--- a/sleepecg/io/sleep_readers.py
+++ b/sleepecg/io/sleep_readers.py
@@ -49,8 +49,8 @@ class SleepRecord:
     sleep_stages : np.ndarray, optional
         Sleep stages according to AASM guidelines, stored as integers as
         defined by `SleepStage`, by default `None`
-    fs_sleep_stages : float, optional
-        Sampling frequency of attribute `sleep_stages`, by default `None`
+    sleep_stage_duration : int, optional
+        Duration of each sleep stage in seconds, by default `None`
     id : str, optional
         The record's ID, by default `None`.
     recording_start_time : datetime.time, optional
@@ -61,7 +61,7 @@ class SleepRecord:
     """
 
     sleep_stages: Optional[np.ndarray] = None
-    fs_sleep_stages: Optional[float] = None
+    sleep_stage_duration: Optional[int] = None
     id: Optional[str] = None
     recording_start_time: Optional[datetime.time] = None
     heartbeat_times: Optional[np.ndarray] = None
@@ -69,7 +69,7 @@ class SleepRecord:
 
 class _ParseNsrrXmlResult(NamedTuple):
     sleep_stages: np.ndarray
-    fs_sleep_stages: float
+    sleep_stage_duration: int
     recording_start_time: datetime.time
 
 
@@ -129,7 +129,7 @@ def _parse_nsrr_xml(xml_filepath: Path) -> _ParseNsrrXmlResult:
 
     return _ParseNsrrXmlResult(
         np.array(annot_stages, dtype=np.int8),
-        1 / epoch_length,
+        epoch_length,
         start_time,
     )
 
@@ -257,7 +257,7 @@ def read_mesa(
 
         yield SleepRecord(
             sleep_stages=parsed_xml.sleep_stages,
-            fs_sleep_stages=parsed_xml.fs_sleep_stages,
+            sleep_stage_duration=parsed_xml.sleep_stage_duration,
             id=record_id,
             recording_start_time=parsed_xml.recording_start_time,
             heartbeat_times=heartbeat_times,
--- a/sleepecg/io/sleep_readers.py
+++ b/sleepecg/io/sleep_readers.py
@@ -13,6 +13,7 @@ from xml.etree import ElementTree
 
 import numpy as np
 
+from ..config import get_config
 from ..heartbeats import detect_heartbeats
 from .nsrr import get_nsrr_url, list_nsrr
 from .utils import download_file
@@ -135,7 +136,7 @@ def _parse_nsrr_xml(xml_filepath: Path) -> _ParseNsrrXmlResult:
 
 
 def read_mesa(
-    data_dir: Union[str, Path],
+    data_dir: Optional[Union[str, Path]] = None,
     records_pattern: str = '*',
     use_preprocessed_heartbeats: bool = True,
     offline: bool = False,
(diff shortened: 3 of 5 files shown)


cbrnr:
How do you want to separate `ecg_readers` from `sleep_readers`? Both must contain ECG, but the latter must have sleep stage annotations?

hofaflo:
> How do you want to separate `ecg_readers` from `sleep_readers`? Both must contain ECG, but the latter must have sleep stage annotations?

Exactly!

hofaflo:
rebased onto main to be able to use `data_dir` from the configuration introduced in #39

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [UNRELEASED]
 ### Added
 - Configuration via a YAML file ([#39](https://github.com/cbrnr/sleepecg/pull/39) by [Florian Hofer](https://github.com/hofaflo))
+- Function for reading data from the [MESA](https://sleepdata.org/datasets/mesa) datasets ([#28](https://github.com/cbrnr/sleepecg/pull/28) by [Florian Hofer](https://github.com/hofaflo))
```
```suggestion
- Add function for reading data from the [MESA](https://sleepdata.org/datasets/mesa) datasets ([#28](https://github.com/cbrnr/sleepecg/pull/28) by [Florian Hofer](https://github.com/hofaflo))
```

sleepecg/io/__init__.py
```diff
@@ -2,11 +2,14 @@
 
 from .ecg_readers import read_gudb, read_ltdb, read_mitdb
 from .nsrr import download_nsrr, set_nsrr_token
+from .sleep_readers import SleepRecord, read_mesa
 
 __all__ = [
+    'SleepRecord',
```
cbrnr, line 8: Why does this need to be exposed? Are users expected to instantiate this class directly? If not and this is handled via corresponding loader functions (which is what I believe), I'd remove this from here.

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+    'SleepStage',
+    'read_mesa',
+]
+
+
+class SleepStage(IntEnum):
+    """
+    Mapping of AASM sleep stages to integers.
+
+    To facilitate hypnogram plotting, values increase with wakefulness.
```
cbrnr, line 32:
```suggestion
    This facilitates hypnogram plotting (values increase with wakefulness).
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+    N1 = 3
+    N2 = 2
+    N3 = 1
+    UNDEFINED = -1
+
+
+@dataclass
+class SleepRecord:
+    """
+    Dataclass to store a single sleep record.
```
cbrnr, line 46:
```suggestion
    Store a single sleep record.
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+
+    Attributes
+    ----------
+    sleep_stages : np.ndarray, optional
+        Sleep stages according to AASM guidelines, stored as integers as
+        defined by `SleepStage`, by default `None`
```
cbrnr, line 52:
```suggestion
        defined by `SleepStage`, by default `None`.
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+    sleep_stage_duration : int, optional
+        Duration of each sleep stage in seconds, by default `None`
```
cbrnr, line 54:
```suggestion
        Duration of each sleep stage in seconds, by default `None`.
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+
+class _ParseNsrrXmlResult(NamedTuple):
+    sleep_stages: np.ndarray
+    sleep_stage_duration: int
+    recording_start_time: datetime.time
+
+
+def _parse_nsrr_xml(xml_filepath: Path) -> _ParseNsrrXmlResult:
+    """
+    Parse NSRR xml sleep stage annotation file.
```
cbrnr, line 79:
```suggestion
    Parse NSRR XML sleep stage annotation file.
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+    xml_filepath : pathlib.Path
+        Path of the annotation file to read.
+
+    Returns
+    -------
+    sleep_stages : np.ndarray
+        Sleep stages according to AASM guidelines, stored as integers as
+        defined by `SleepStage`.
+    fs_sleep_stages : float
+        Sampling frequency of `sleep_stages`.
```
cbrnr, line 92: This should be `sleep_stage_duration: int` I guess.

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+        'Stage 1 sleep|1': SleepStage.N1,
+        'Stage 2 sleep|2': SleepStage.N2,
+        'Stage 3 sleep|3': SleepStage.N3,
+        'Stage 4 sleep|4': SleepStage.N3,
+        'REM sleep|5': SleepStage.REM,
+        'Unscored|9': SleepStage.UNDEFINED,
+    }
+
+    tree = ElementTree.parse(xml_filepath)
+    root = tree.getroot()
```
cbrnr, line 108:
```suggestion
    root = ElementTree.parse(xml_filepath).getroot()
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+
+def read_mesa(
+    data_dir: Optional[Union[str, Path]] = None,
+    records_pattern: str = '*',
+    use_preprocessed_heartbeats: bool = True,
+    offline: bool = False,
+    persist_edfs: bool = False,
+) -> Iterator[SleepRecord]:
+    """
+    Lazily reads records from MESA (https://sleepdata.org/datasets/mesa).
```
cbrnr, line 146:
```suggestion
    Lazily read records from MESA (https://sleepdata.org/datasets/mesa).
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+
+    Each MESA record consists of a `.edf` files containing raw
```
cbrnr, line 148:
```suggestion
    Each MESA record consists of a `.edf` file containing raw
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+    polysomnography data and an `.xml` files containing annotated events.
```
cbrnr, line 149:
```suggestion
    polysomnography data and an `.xml` file containing annotated events.
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+    Since the entire MESA dataset requires about 385 GB of disk space,
+    `.edf` files can be deleted after the heartbeat times have been
```
cbrnr, line 151:
```suggestion
    `.edf` files can be deleted after heartbeat times have been
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+    extracted. Heartbeat times are stored to a `.npy` file in
```
cbrnr, line 152:
```suggestion
    extracted. Heartbeat times are cached in a `.npy` file in
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+    `<data_dir>/mesa/preprocessed/heartbeats`.
+
+    Parameters
+    ----------
+    data_dir : str | pathlib.Path
```
cbrnr, line 157:
```suggestion
    data_dir : str | pathlib.Path, optional
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+        Directory where all datasets are stored. If `None` (default), the
+        value will be taken from the configuration.
+    records_pattern : str, optional
+         Glob-like pattern to select record IDs, by default `'*'`.
+    use_preprocessed_heartbeats : bool, optional
```
cbrnr, line 162:
```suggestion
    use_cached_heartbeats : bool, optional
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+        np.array(annot_stages, dtype=np.int8),
+        epoch_length,
+        start_time,
+    )
+
+
+def read_mesa(
+    data_dir: Optional[Union[str, Path]] = None,
+    records_pattern: str = '*',
+    use_preprocessed_heartbeats: bool = True,
```
cbrnr, line 141:
```suggestion
    use_cached_heartbeats: bool = True,
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+
+    Parameters
+    ----------
+    data_dir : str | pathlib.Path
+        Directory where all datasets are stored. If `None` (default), the
+        value will be taken from the configuration.
+    records_pattern : str, optional
+         Glob-like pattern to select record IDs, by default `'*'`.
+    use_preprocessed_heartbeats : bool, optional
+        If `True`, get heartbeat times directly from the stored `.npy`
```
cbrnr, line 163:
```suggestion
        If `True`, get heartbeat times directly from the cached `.npy`
```

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+        file, so `.edf` files are only downloaded for records which have
+        not yet been preprocessed. By default `True`.
+    offline : bool, optional
+        If `True`, search for local files only instead of using the NSRR
+        API, by default `False`.
+    persist_edfs : bool, optional
```
cbrnr, line 169: Can we rename this parameter to `keep_edfs`? I think this is easier to understand.

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+        If `False`, remove `.edf` after heartbeat detection, by default
+        `False`.
+
+    Yields
+    ------
+    SleepRecord
+        Each element in the generator is a `SleepRecord`, containing
+        heartbeat times, sleep stages, sleep stage sampling frequency,
+        record ID and record start time.
```
cbrnr, line 178: This is inconsistent with the actual `SleepRecord` class, which doesn't contain the sampling frequency. Also, I would list the attributes in the same order as they appear in the class definition. Or you can also remove the details entirely (everything after the first comma).

sleepecg/io/sleep_readers.py
```diff
@@ -0,0 +1,269 @@
...
+        checksums.update(edf_files)
+    else:
+        xml_files = sorted(annotations_dir.glob(f'mesa-sleep-{records_pattern}-nsrr.xml'))
+        requested_records = [file.stem[:-5] for file in xml_files]
+        if not use_preprocessed_heartbeats:
+            edf_files = sorted(edf_dir.glob(f'mesa-sleep-{records_pattern}.edf'))
+
+    for record_id in requested_records:
+        heartbeats_file = heartbeats_dir / f'{record_id}.npy'
+        if use_preprocessed_heartbeats and heartbeats_file.is_file():
```
cbrnr, line 227:
```suggestion
        if use_cached_heartbeats and heartbeats_file.is_file():
```

cbrnr (commented):
Looks good besides some minor suggestions. It is hard for me to debug this (verify that it works) without a test - did you plan to add some tests?

sleepecg/io/sleep_readers.py
hofaflo: I'd leave this as it is, since it should explain why the values increase with wakefulness, not why the enum is there in the first place (which is to avoid having "magic" numeric literals in the reader functions). Or reformulate to make the intention clearer, e.g.
```suggestion
    The values increase with wakefulness, so the order of stages will be
    correct in a hypnogram plot.
```

sleepecg/io/sleep_readers.py
hofaflo:
```suggestion
    sleep_stage_duration : int
        Duration of each sleep stage in seconds.
```

sleepecg/io/__init__.py
hofaflo: True... I'm still not completely clear on what's supposed to be in `__all__`, depending on the file (`__init__.py`/`some_module.py`) 😅. Would you keep exposing `SleepStage` and `SleepRecord` in `io.sleep_readers`? They shouldn't be used by a package user, but will be required by other modules (like e.g. `download_file` in `io.utils`).

sleepecg/io/sleep_readers.py
hofaflo: I'll take the last option

hofaflo:
> did you plan to add some tests?

MESA (like almost every [NSRR dataset](https://sleepdata.org/datasets/)) requires a token for file access, so there's little to be tested without that token :/

sleepecg/io/__init__.py
cbrnr: Then I wouldn't expose it in `__all__`, which I interpret as the public API (but I might be wrong). Internal modules can always import whatever they want from other local modules.

sleepecg/io/sleep_readers.py
cbrnr: Ah, sorry, I misread that. Then keep the original please.

cbrnr:
> MESA (like almost every NSRR dataset) requires a token for file access, so there's little to be tested without that token :/

Hm. And the token is just for personal use I assume? Can you provide a little example script that I could try to see if the function works as expected?

hofaflo:
> Hm. And the token is just for personal use I assume? 

Exactly :/

> Can you provide a little example script that I could try to see if the function works as expected?

I could create some edf and xml files having the same structure... then you could test everything besides the download 🤔

cbrnr:
> I could create some edf and xml files having the same structure... then you could test everything besides the download 🤔

This would be a nice test I think. Regarding the token, how does this work? Do you have instructions? Should we include a short how-to?

hofaflo:
> This would be a nice test I think. 

👍 

> Regarding the token, how does this work? Do you have instructions? Should we include a short how-to?

You create an account at sleepdata.org to get a token - to make that token work for a specific dataset, you fill out a form stating which dataset(s) you want to use and for what, then wait for approval. I'll create a documentation page regarding datasets/readers in the future and include a how-to there.

sleepecg/io/__init__.py
hofaflo: Okay, works for me! I'll open a PR to standardize this once the current ones are merged :)

sleepecg/io/sleep_readers.py
```diff
@@ -139,16 +137,16 @@ def read_mesa(
...
     Lazily read records from MESA (https://sleepdata.org/datasets/mesa).
 
-    Each MESA record consists of a `.edf` file containing raw
+    Each MESA record consists of an `.edf` file containing raw
     polysomnography data and an `.xml` file containing annotated events.
     Since the entire MESA dataset requires about 385 GB of disk space,
-    `.edf` files can be deleted after heartbeat times have been
-    extracted. Heartbeat times are cached in a `.npy` file in
+    `.edf` files can be deleted after heartbeat times have been extracted.
+    Heartbeat times are cached in a `.npy` file in
```
cbrnr, line 149:
```suggestion
    Heartbeat times are cached in an `.npy` file in
```

cbrnr (commented):
Nice tests 👍! Only one minor nitpick, since you use `np.random` I think the function(s) should have a random state parameter to be reproducible.

cbrnr:
Do you want to add a default value and type annotation for `random_state`?

hofaflo:
> Do you want to add a default value and type annotation for `random_state`?

`np.random.default_rng` accepts `Union[None, _ArrayLikeInt_co, SeedSequence, BitGenerator, Generator]` which I wanted to avoid, but maybe we could only use `int`?

cbrnr:
Yes, let's just support an `int`!

cbrnr:
Great! Ready to merge then?

hofaflo:
Yep :)

PR #32, opened 2021-08-24
Separate reader functions per database

### LINKED ISSUE #30: Separate reader functions per database
Currently, we have `read_mitbih()`, which reads two specific data sets from the MIT-BIH database: MITDB and LTDB. It would be useful to have a dedicated reader function for each database, i.e. `read_mitdb()` and `read_ltdb()` (both of which could use  a private version of `read_mitbih()` under the hood.

hofaflo:
closes #30


cbrnr:
Thanks @hofaflo!

PR #33, opened 2021-08-24
Use `~/.sleepecg/datasets` as default `data_dir` for readers

hofaflo:
I've introduced a "global" variable `sleepecg._globals._DATA_DIR` to single-source `~/.sleepecg/datasets`. At the moment (3f2f45b), only the public `read_*` functions use this as a default. I would not extend this to the physionet- and nsrr-download functions since `download_physionet` is only used internally and I'd prefer `download_nsrr` to mirror the behaviour of the NSRR Ruby Gem (i.e. download to CWD by default).


cbrnr:
I think it's OK to apply the default location only to public reader functions. @hofaflo can you rebase please?

cbrnr:
At some point we could also think about storing such configuration in `~/.sleepecg/config.ini` or something (which overrides global defaults, which users could set/read with `set_config()` and `get_config()` or something along these lines).

cbrnr:
I wonder if we should remove the underscore from `globals.py` and `DATA_DIR` (they are kind of public and it would be more readable).

cbrnr:
Do you also want to update the example and the benchmark script?

hofaflo:
Oh right, thanks! I'd leave the benchmark-script as-is, since the value is specified in the readme as well where it can't be retrieved from `globals.py`.

cbrnr:
Thanks @hofaflo!

PR #39, opened 2021-11-16
Add configuration interface

hofaflo:
This adds the possibility to store and retrieve configuration values which are stored under `~/.sleepecg/config.yml`, with default values sourced from `site-packages/sleepecg/config.yml`. The module `globals` is made obsolete and removed.

### DIFF
--- a/sleepecg/config.py
+++ b/sleepecg/config.py
@@ -61,11 +61,10 @@ def get_config(key: Optional[str] = None) -> Any:
 
     if key is None:
         return config
-    try:
-        return config[key]
-    except KeyError:
+    if key not in config:
         options = ', '.join(config)
-        raise ValueError(f'Trying to get invalid config key: {key!r}, possible options: {options}') from None  # noqa: E501
+        raise ValueError(f'Trying to get invalid config key: {key!r}, possible options: {options}')  # noqa: E501
+    return config[key]
 
 
 def set_config(**kwargs):
--- a/sleepecg/config.py
+++ b/sleepecg/config.py
@@ -25,6 +25,8 @@ def _read_yaml(path: Path) -> Dict[str, Any]:
         # empty .yml-files are loaded as `None`
         if cfg is None:
             return {}
+        if not isinstance(cfg, dict):
+            raise ValueError(f'Invalid YAML config file at {_USER_CONFIG_PATH}')
         return cfg
 
 
@@ -49,7 +51,11 @@ def get_config(key: Optional[str] = None) -> Any:
     """
     config = _read_yaml(_DEFAULT_CONFIG_PATH)
     with contextlib.suppress(FileNotFoundError):
-        config.update(_read_yaml(_USER_CONFIG_PATH))
+        user_config = _read_yaml(_USER_CONFIG_PATH)
+        for key in user_config:
+            if key not in config:
+                raise ValueError(f'Invalid key found in user config at {_USER_CONFIG_PATH}: {key}')  # noqa: E501
+        config.update(user_config)
 
     if key is None:
         return config
--- a/sleepecg/config.py
+++ b/sleepecg/config.py
@@ -94,5 +94,6 @@ def set_config(**kwargs):
         else:
             user_config[key] = value
 
+    _USER_CONFIG_PATH.parent.mkdir(parents=True, exist_ok=True)
     with open(_USER_CONFIG_PATH, 'w') as user_config_file:
         yaml.dump(user_config, user_config_file)
--- a/sleepecg/config.py
+++ b/sleepecg/config.py
@@ -22,14 +22,15 @@ def _read_yaml(path: Path) -> Dict[str, Any]:
     try:
         with open(path) as file:
             cfg = yaml.safe_load(file)
-            # empty .yml-files are loaded as `None`
-            if cfg is None:
-                return {}
-            if not isinstance(cfg, dict):
-                raise ValueError(f'Invalid YAML config file at {_USER_CONFIG_PATH}')
-            return cfg
     except FileNotFoundError:
         return {}
+
+    # empty .yml-files are loaded as `None`
+    if cfg is None:
+        return {}
+    if not isinstance(cfg, dict):
+        raise ValueError(f'Invalid YAML config file at {path}')
+    return cfg
 
 
 def get_config(key: Optional[str] = None) -> Any:
--- a/sleepecg/config.py
+++ b/sleepecg/config.py
@@ -4,7 +4,6 @@
 
 """Functions for getting and setting configuration values."""
 
-import contextlib
 from pathlib import Path
 from typing import Any, Dict, Optional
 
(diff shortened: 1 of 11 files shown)


hofaflo:
ready @cbrnr :)

sleepecg/config.py
```diff
@@ -0,0 +1,98 @@
...
+        settings are returned in a dictionary, by default `None`.
+
+    Returns
+    -------
+    typing.Any
+        The configuration value.
+    """
+    config = _read_yaml(_DEFAULT_CONFIG_PATH)
+    with contextlib.suppress(FileNotFoundError):
+        config.update(_read_yaml(_USER_CONFIG_PATH))
```
cbrnr, line 52: Why do you need `contextlib.suppress` here instead of `try/except FileNotFoundError`?

sleepecg/config.py
```diff
@@ -0,0 +1,98 @@
...
+
+    if key is None:
+        return config
+    try:
+        return config[key]
+    except KeyError:
+        options = ', '.join(config)
+        raise ValueError(f'Trying to get invalid config key: {key!r}, possible options: {options}') from None  # noqa: E501
```
cbrnr, line 60: Why do you `raise from None` here? In `set_config` you are just raising the `ValueError`, which I'd also do here.

cbrnr (commented):
Looks good. Can you add one or two tests to test some basic config functionality?

sleepecg/config.py
hofaflo: I wanted to avoid using `except FileNotFoundError: pass`... as a third option, I could check if the file exists before reading?

sleepecg/config.py
hofaflo: being inside an `except`-block, this `raise` would let the traceback include the original `KeyError` which might be confusing

hofaflo:
added some tests and basic config file validation

sleepecg/config.py
cbrnr: What about putting the `try/except` into `_read_yaml` then?

sleepecg/config.py
cbrnr: And what about checking with `if key in config`?

sleepecg/test/test_config.py
```diff
@@ -0,0 +1,54 @@
...
+    assert get_config('data_dir') == '~/.sleepecg/datasets'
+    set_config(data_dir='some_dir')
+    assert get_config('data_dir') == 'some_dir'
+    set_config(data_dir=None)
+    assert get_config('data_dir') == '~/.sleepecg/datasets'
+
+
+def test_set_invalid_config():
+    """Test trying to set an invalid configuration key."""
+    with pytest.raises(ValueError):
```
cbrnr, line 41: Maybe include some snippets of the actual error message to be more specific?

sleepecg/test/test_config.py
```diff
@@ -0,0 +1,54 @@
...
+        set_config(invalid_key='some_value')
+
+
+def test_get_invalid_config():
+    """Test trying to get an invalid configuration key."""
+    with pytest.raises(ValueError):
```
cbrnr, line 47: Same as above.

sleepecg/test/test_config.py
```diff
@@ -0,0 +1,54 @@
...
+"""Tests for the configuration interface."""
+
+from pathlib import Path
+
+import pytest
+
+import sleepecg.config
+from sleepecg import get_config, set_config
+
+_TESTCONFIG_PATH = '~/.sleepecg/testdata/testconfig.yml'
```
cbrnr, line 14: I guess it's OK to test in the user directory, but we should make sure that any directory or file that gets created during testing is always deleted (even if a test fails). I'm not sure if this is currently guaranteed. I thought about using the `tmp_path` fixture for that, but then we can't test the default location (which is in the user's home) - but maybe we don't have to?

sleepecg/test/test_config.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+@pytest.fixture(autouse=True)
+def temp_test_config():
+    """Create, use and delete a temporary user config file for testing."""
+    user_config_path_backup = sleepecg.config._USER_CONFIG_PATH
+    sleepecg.config._USER_CONFIG_PATH = Path(_TESTCONFIG_PATH).expanduser()
+
+    yield
```
cbrnr, line 23: What is this? I guess it's related to the `autouse=True`, but can you quickly explain what's happening?

sleepecg/test/test_config.py
hofaflo: Yes, good idea! I agree, the default location should not be necessary

sleepecg/test/test_config.py
hofaflo: It's a general syntax for pytest fixtures, the `yield` is a placeholder for the actual testcase.
From [the pytest API reference](https://docs.pytest.org/en/6.2.x/reference.html#pytest-fixture):
"Fixtures can provide their values to test functions using return or yield statements. When using yield the code block after the yield statement is executed as teardown code regardless of the test outcome, and must yield exactly once."

I'll add comments to explain :)

cbrnr: pytest is like black magic. So this means that `autouse=True` is the value that's provided to the test here?

hofaflo: :D
no, autouse just lets all tests in this module use this fixture by default (instead of specifying the fixture as a parameter in each function definition)

cbrnr: So what happens in this specific test in the `yield` line then?

sleepecg/config.py
```diff
@@ -20,14 +19,17 @@
...
+        with open(path) as file:
+            cfg = yaml.safe_load(file)
+            # empty .yml-files are loaded as `None`
+            if cfg is None:
+                return {}
+            if not isinstance(cfg, dict):
+                raise ValueError(f'Invalid YAML config file at {_USER_CONFIG_PATH}')
+            return cfg
+    except FileNotFoundError:
+        return {}
```
cbrnr, line 32: Hm, in this case there's a lot going on in the `try` block, which maybe means that it would be better to just check if the file exists first?

hofaflo: oh right, most of that doesn't even need to be inside the `with`, fixed in 327a978:)

sleepecg/config.py
```diff
@@ -22,15 +22,16 @@ def _read_yaml(path: Path) -> Dict[str, Any]:
...
-            if not isinstance(cfg, dict):
-                raise ValueError(f'Invalid YAML config file at {_USER_CONFIG_PATH}')
-            return cfg
     except FileNotFoundError:
         return {}
 
+    # empty .yml-files are loaded as `None`
+    if cfg is None:
+        return {}
+    if not isinstance(cfg, dict):
```
cbrnr, line 31: What on earth gets returned from an invalide YAML file?

hofaflo: If it's actually syntactically invalid, PyYAML will raise a `ParserError`... but a syntactically valid YAML file might contain a list/str/int/float instead of a mapping, which wouldn't make too much sense for our usecase^^

sleepecg/test/test_config.py
hofaflo: Oh, `temp_test_config` is not a test, but a fixture... it defines the setup and cleanup steps used for the actual testcases.

cbrnr: Ah, got it!

sleepecg/config.py
cbrnr: Right, I guess that's OK then.

sleepecg/config.py
cbrnr: So are you going to change it or leave it as is 🤔?

hofaflo: right, sorry, fixed!

cbrnr:
Perfect! Thanks @hofaflo!

PR #41, opened 2021-12-01
Build only 64bit wheels

hofaflo:


cbrnr:
Could it be that we run cibuildwheel on the three platforms, and we actually build wheels for all three platforms three times?

hofaflo:
We build wheels for 3.7, 3.8 and 3.9 on each of the three platforms... do you mean that?

cbrnr:
OK, this makes sense. No, I thought we were building wheels for all platforms on each platform.

Do you know how to access the artifacts uploaded in each job? For example, the output on `macos-10.15` says that 5 files were uploaded. Where are these files? And why 5?

cbrnr:
OK, I see the produced wheels now, but still don't know where to download them.

Related question: why are there 6 wheels for Linux? 5 for macOS and 3 for Windows make sense.

Another idea: we could drop support for Python 3.7 and add 3.10 instead.

cbrnr:
> manylinux1 and manylinux2014... but apparently only one wheel is built for each python version and then duplicated with a different name 🤷

According to https://github.com/pypa/manylinux/ these should be different builds though.

> Fine with me, though I'd wait until our dependencies have wheels for 3.10 😅.

Definitely. Which packages are currently missing?

hofaflo:
> According to https://github.com/pypa/manylinux/ these should be different builds though.

Strange 🤔 

> Definitely. Which packages are currently missing?

I just checked (on windows), the core dependencies seem to work, but numba (used for heartbeat detection) and pyedflib (used for testing `read_mesa`) don't.

cbrnr:
Thanks @hofaflo!

Numba should be available soon (https://github.com/numba/numba/issues/7562). Not sure about pyedflib though, we could ask, but let's just wait a couple of weeks. I'll create an issue so that we don't forget.

Not sure what's up with manylinux, I'll create a separate issue so that I can merge this PR.

PR #42, opened 2021-12-01
Fix failing readthedocs build

hofaflo:
Building the docs on readthedocs is currently [failing](https://readthedocs.org/projects/sleepecg/builds/15421420/) because it attempts to install numpy 1.22.0rc1, which requires python 3.8. Using `pip` as the package installation method (instead of `setuptools`) and placing the doc requirements into an extras-section in `setup.cfg` (instead of `doc/requirements-doc.txt`) appears to fix that (see [this test build](https://readthedocs.org/projects/test-sleepecg/builds/15421560/)).

### DIFF
--- a/doc/requirements-doc.txt
+++ b/doc/requirements-doc.txt
@@ -1,4 +1,0 @@
-myst-parser
-numpydoc
-pydata_sphinx_theme
-sphinx
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -4,13 +4,18 @@
 
 version: 2
 
+build:
+  os: ubuntu-20.04
+  tools:
+    python: "3.7"
+
 sphinx:
   configuration: doc/source/conf.py
 
 python:
-  version: 3.7
   system_packages: true
   install:
-  - requirements: doc/requirements-doc.txt
-  - method: setuptools
+  - method: pip
     path: "."
+    extra_requirements:
+      - doc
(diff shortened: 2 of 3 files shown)


cbrnr:
Looks good!

- Should we use Python 3.8 here?
- Could we use readthedocs to build and render our docs on each PR instead of CircleCI?

hofaflo:
> Should we use Python 3.8 here?

It's installing a 3.7-compatible numpy now, so I think we can leave it as it is.

> Could we use readthedocs to build and render our docs on each PR instead of CircleCI?

Looks like yes... I enabled it in the RTD settings, should work with the next commit.
Thanks for the idea, good thing I haven't started with CircleCI yet 😅

hofaflo:
Confirmed in #43: https://readthedocs.org/projects/sleepecg/builds/15422568/ 🥳

cbrnr:
Wait, what's going on with the RTD build?

1. It's currently failing in #43 (Python >= 3.8)
2. How did the RTD check magically appear in #43? 😕

hofaflo:
> 1. It's currently failing in [Execute example scripts with pytest #43](https://github.com/cbrnr/sleepecg/pull/43) (Python >= 3.8)

Right, that will be fixed by this PR

> 2. How did the RTD check magically appear in [Execute example scripts with pytest #43](https://github.com/cbrnr/sleepecg/pull/43)? 😕

The "action" is activated by a [checkbox in the RTD settings](https://docs.readthedocs.io/en/stable/pull-requests.html)

cbrnr:
Thanks @hofaflo! Let's see how this works!

PR #43, opened 2021-12-01
Execute example scripts with pytest

hofaflo:
To make sure the example scripts work... if an example should not be tested (e.g. because takes a long time, like the benchmark), it can be excluded by adding its name to `EXCLUDE` in `sleepecg/test/test_examples.py`.

### DIFF
--- a/sleepecg/test/test_examples.py
+++ b/sleepecg/test/test_examples.py
@@ -1,4 +1,3 @@
-# %%
 # Authors: Florian Hofer
 #
 # License: BSD (3-clause)
--- a/sleepecg/test/test_examples.py
+++ b/sleepecg/test/test_examples.py
@@ -1,0 +1,32 @@
+# Authors: Florian Hofer
+#
+# License: BSD (3-clause)
+
+"""Tests to make sure examples don't crash."""
+
+import runpy
+from pathlib import Path
+
+import matplotlib.pyplot as plt
+import pytest
+
+EXCLUDE = [
+    '*/benchmark/*',
+]
+
+examples_dir = (Path(__file__).parent / '../../examples').resolve()
+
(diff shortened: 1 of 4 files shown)


sleepecg/test/test_examples.py
```diff
@@ -0,0 +1,32 @@
...
+    '*/benchmark/*',
+]
+
+examples_dir = (Path(__file__).parent / '../../examples').resolve()
+
+files_to_test = []
+for pyfile in examples_dir.rglob('*.py'):
+    for pattern in EXCLUDE:
+        if not pyfile.match(pattern):
+            files_to_test.append(pyfile)
```
cbrnr, line 23: The logic doesn't work out - if a file does not match the first pattern you are already adding it to the files to test. But what if it matches another pattern defined in `EXCLUDE`?

hofaflo: True, fixed!

cbrnr:
Can you rebase so that we can see if the docs build?

hofaflo:
Done, they do!

cbrnr:
Wow, I'm actually surprised I didn't have this idea earlier! This is super convenient and we don't need another service. Just a quick reminder, RTD is configured to show "stable" (latest release) and "latest" (HEAD), right?

hofaflo:
Right, with "stable" being the default... although the `docs|passing`-badge in the main readme links to "latest". Should we change that?

cbrnr:
Good question. When viewing the README on GitHub we are looking at the "latest" version, so the badge should point to the same version. Therefore, I guess it's OK?

hofaflo:
Okay!

cbrnr:
Thanks @hofaflo!

PR #46, opened 2021-12-01
Make public API exports less redundant

hofaflo:
To avoid repeating module member names in `__init__.py` files, I'd suggest the following guidelines:
- Every non-public member (i.e. every member not intended to be accessed by an end user) is prefixed with an underscore: `_`.
- Modules export their public members explicitly (i.e. alphabetically list them in `__all__`). Otherwise their imports would be exported as well.
- Inside a (sub-)package's `__init__.py`, `__all__` is _not_ set and module members are imported using `from .module import *`. This avoids listing member names multiple times.

Notes:
- `SleepRecord` and `ECGRecord` will usually not be instantiated by a user, but should be explained in the docs as they are yielded by public generators. Only public members (i.e. those not starting with an underscore) are listed by Sphinx' autosummary. Therefore these two need to have a public name, but are not exported.
- We could have tests to make sure these guidelines are followed, WDYT?

### DIFF
--- a/sleepecg/io/sleep_readers.py
+++ b/sleepecg/io/sleep_readers.py
@@ -15,15 +15,15 @@ import numpy as np
 
 from ..config import get_config
 from ..heartbeats import detect_heartbeats
-from .nsrr import get_nsrr_url, list_nsrr
-from .utils import download_file
+from .nsrr import _get_nsrr_url, _list_nsrr
+from .utils import _download_file
 
 __all__ = [
     'read_mesa',
 ]
 
 
-class SleepStage(IntEnum):
+class _SleepStage(IntEnum):
     """
     Mapping of AASM sleep stages to integers.
 
@@ -47,7 +47,7 @@ class SleepRecord:
     ----------
     sleep_stages : np.ndarray, optional
         Sleep stages according to AASM guidelines, stored as integers as
-        defined by `SleepStage`, by default `None`.
+        defined by `_SleepStage`, by default `None`.
     sleep_stage_duration : int, optional
         Duration of each sleep stage in seconds, by default `None`.
     id : str, optional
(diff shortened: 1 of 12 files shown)


cbrnr:
Yes, I think this is a good approach! The alternative is to not use `__all__` at all and import public functions/classes/... explicitly in the various `__init__.py` files, which might be a bit more flexible. I don't have a strong opinion, so I'll let you decide how to proceed!

If we go with your suggestions, we don't even have to be super strict with our names. For example, we could continue to use names without an underscore prefix in certain cases even if they are not public (e.g. the various constants only used within a module), because they will not get exported automatically anyway! WDYT?

hofaflo:
I actually like your approach better, because everything is in one spot (and it avoids `*` imports) :D

sleepecg/io/sleep_readers.py
```diff
@@ -15,15 +15,11 @@
...
+from .nsrr import _get_nsrr_url, _list_nsrr
+from .utils import _download_file
 
-__all__ = [
-    'read_mesa',
-]
 
-
-class SleepStage(IntEnum):
+class _SleepStage(IntEnum):
```
cbrnr, line 22: I'd actually be OK with leaving this as `SleepStage`. We explicitly import objects in `__init__.py` anyway, so we don't always have to be strict with our naming. Same for some constants that look weird with an underscore. Don't users at least see `SleepStage` objects sometimes because they get returned? But I'll leave it up to you.

hofaflo: Users don't see it, since `SleepRecord.sleep_stages` is an integer array, but it would make sense to have it in the documentation, thanks for noticing!

I'd prefer to keep underscore prefixes for constants though.

cbrnr:
Thanks @hofaflo!

PR #48, opened 2021-12-07
Add reader for SHHS

hofaflo:
[SHHS](https://sleepdata.org/datasets/shhs)

This is pretty similar to `read_mesa`, except for the following differences:
- SHHS is grouped into two "visits" (there are two recorded nights for some participants), therefore there is an additional directory level (shhs1/shhs2).
- For MESA the ECG channel name is "EKG", for SHHS it's "ECG".
- SHHS does not include actigraphy data (while `read_mesa` doesn't read that data yet, this is planned).

Due to the similarity, large parts of `read_mesa` and `read_shhs` could be put into a new function. I'd like to postpone refactoring this until `read_mesa` handles actigraphy data and both functions handle metadata.

### DIFF
--- a/sleepecg/io/nsrr.py
+++ b/sleepecg/io/nsrr.py
@@ -119,6 +119,42 @@ def _list_nsrr(
     return files
 
 
+def _download_nsrr_file(
+    url: str,
+    target_filepath: Path,
+    checksum: str,
+):
+    """
+    Download a file from `url` to `target_filepath` and verify `checksum`.
+
+    This is a wrapper around `sleepecg.io.utils._download_file` to provide
+    a helpful error message in case the currently set token does not grant
+    access to the requested file.
+
+    Parameters
+    ----------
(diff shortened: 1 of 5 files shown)


sleepecg/io/nsrr.py
```diff
@@ -119,6 +119,42 @@ def _list_nsrr(
     return files
 
 
+def _download_nsrr_file(
```
cbrnr, line 122: Is the only difference to `_download_file` the missing `checksum_type` parameter?

hofaflo: Almost - in case the checksum doesn't match, this function checks whether the downloaded "file" could be the files-overview page on sleepdata.org (which the request is redirected to in case the token is not valid for the requested file), to give a more helpful error message than "checksum mismatch".

cbrnr:
Thanks @hofaflo!

PR #49, opened 2021-12-12
Use latest cibuildwheel version 2.3.0

cbrnr:
Let's see how this affects the number of wheels that are built.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,17 +7,6 @@ on:
     # allow manual runs on branches without a PR
 
 
-env:
-  CIBW_TEST_REQUIRES: pytest
-  CIBW_TEST_EXTRAS: full
-  CIBW_TEST_COMMAND: pytest {package}
-  CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
-  CIBW_SKIP: pp*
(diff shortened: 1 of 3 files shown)


cbrnr:
Numba for Python 3.10 should be available very soon hopefully (https://github.com/numba/numba/issues/7562), so let's restart the tests once this is the case.

hofaflo:
Nice idea!

Skipping 3.10 for now by extending `CIBW_SKIP: pp*` to `CIBW_SKIP: pp* cp310-*` should allow to test this right now.

Side note: using this newer cibuildwheel version would allow configuration via `pyproject.toml`, so we could avoid setting the same environment variables in both workflows.

cbrnr:
OK! Can you push your suggested changes?

hofaflo:
cibuildwheel v2.2.0 introduced support for building musllinux wheels, which [doesn't work in our case](https://github.com/cbrnr/sleepecg/runs/4499217025?check_suite_focus=true), so I guess we'll skip them?

Yep, looks like your goal was achieved, there's now only one manylinux wheel (which is tagged for multiple manylinux versions) per python version 🥳

cbrnr:
Yes, I think we can safely ignore musl.

hofaflo:
Perfect, thanks @cbrnr!

PR #54, opened 2021-12-16
Skip test workflow on doc-only changes

hofaflo:
In case a commit/PR only changes files in `doc`, the build/test workflow doesn't need to run. This was already intended, but didn't happen due to a typo.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ on:
       - main
   pull_request:
     paths-ignore:
-      - 'docs/**'
+      - 'doc/**'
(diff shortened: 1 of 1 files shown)


cbrnr:
Does this automatically work or do we need to use a special commit message or PR title?

hofaflo:
Should work automatically according to [this](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-ignoring-paths)

cbrnr:
Very nice, I didn't know that! Thanks @hofaflo!

PR #56, opened 2021-12-18
Export `sleepecg.io` members to main package 

hofaflo:
To avoid situations like this:

```python
from sleepecg import extract_features
from sleepecg.io import read_mesa
```

I'd like to reduce the API to a single level, so one can do this:

```python
from sleepecg import extract_features, read_mesa
```

Also, to have `SleepRecord`, `SleepStage` and `ECGRecord` included and correctly linked in the docs, they are exported as well.


cbrnr:
Thanks @hofaflo!

PR #57, opened 2021-12-21
Fix mistakes in `feature_extraction`

hofaflo:
While working on the classification pipeline, I found 3 issues in the feature extraction code:

- The column indices used for selecting the requested features is not in the correct order, so the `feature_ids` returned by `extract_features` does not match the order of columns in the returned feature matrices.
- The calculation of `SD1` is missing a square root (see [hrvanalysis](https://github.com/Aura-healthcare/hrv-analysis/blob/master/hrvanalysis/extract_features.py#L539) and [Eq. 8 in this paper](https://psicolibra.it/wp-content/uploads/2013/10/nonlinear_features_of_heart_rate_variability.pdf) for reference).
- If a `SleepRecord` contains annotated `sleep_stages`, the feature matrix is missing 1 row. The subtraction in line 532 was originally intended to avoid having undefined sleep stages at the signal end in case the sleep stages are upsampled (e.g. with original sleep stage sample times `0, 30, 60`, resampling to 10s stages would yield sample times  `0, 10, 20, 30, 40, 50, 60, 70, 80`, so `interp1d` would raise a `ValueError` for the last two stages). Passing a `fill_value` to `interp1d` solves this issue without cutting away the last stage.

### DIFF
--- a/sleepecg/feature_extraction.py
+++ b/sleepecg/feature_extraction.py
@@ -506,7 +506,7 @@ def _extract_features_single(
     rri_required = 'hrv-time' in required_groups or 'hrv-frequency' in required_groups
 
     if record.sleep_stages is not None and record.sleep_stage_duration is not None:
-        record_duration = (len(record.sleep_stages) - 1) * record.sleep_stage_duration
+        record_duration = len(record.sleep_stages) * record.sleep_stage_duration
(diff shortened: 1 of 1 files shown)


cbrnr:
Thanks @hofaflo!

PR #58, opened 2022-01-08
Improve comments in heartbeats.py

cbrnr:
Mostly very minor changes to make the comments more consistent and easier to follow.

### DIFF
--- a/sleepecg/heartbeats.py
+++ b/sleepecg/heartbeats.py
@@ -625,7 +625,8 @@ def _thresholding_py(
                     threshold_F1 /= 2
                     threshold_I1 /= 2
 
-            # a signal peak has been found, so perform a searchback
+            # A signal peak has been found, so performing a searchback
+            # makes sense.
             do_searchback = True
 
             # previous peak index is required to calculate the RR interval
--- a/sleepecg/heartbeats.py
+++ b/sleepecg/heartbeats.py
@@ -47,7 +47,7 @@ def detect_heartbeats(ecg: np.ndarray, fs: float, backend: str = 'c') -> np.ndar
     - The signal might start during a peak, in which case it might have a
       relatively high initial amplitude, which will mess up threshold
       initialization. Therefore, everything until the first zero-crossing
-      is set to 0.
+      is set to zero.
     - The integration window is centered on the filtered signal, i.e. a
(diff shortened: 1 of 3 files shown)


cbrnr:
@hofaflo do you know what's up with the `ImportError: numpy.core.multiarray failed to import`?

hofaflo:
Not sure what the reason is exactly, but I've found out the following things:
The wheel is built using the most recent version of numpy on each Python version (1.21 for 3.7, 1.22 for 3.8/3.9). For testing, numba is installed, which will support NumPy 1.21 only on its next release, so NumPy is downgraded to 1.20.x. The message `RuntimeError: module compiled against API version 0xf but this version of numpy is 0xe` is probably there because of some incompatible change in the NumPy C-API from 1.21 to 1.22. Limiting the NumPy version used for building the wheels in `pyproject.toml` to something below 1.22 (e.g. `'numpy<=1.21.5'`) fixes the issue for Python 3.8 on Windows.

sleepecg/heartbeats.py
```diff
@@ -626,11 +625,10 @@ def _thresholding_py(
                     threshold_F1 /= 2
                     threshold_I1 /= 2
 
-            # A signal peak has been found, so performing a searchback
-            # makes sense.
+            # a signal peak has been found, so perform a searchback
```
hofaflo, line 628: `do_searchback` being `True` is just a necessary requirement for performing a searchback, not a sufficient one (see line 452). 
```suggestion
            # A signal peak has been found, so performing a searchback
            # makes sense.
```

cbrnr:
> Limiting the NumPy version used for building the wheels in pyproject.toml to something below 1.22 (e.g. 'numpy<=1.21.5') fixes the issue for Python 3.8 on Windows.

Do you want to push this as a hotfix until Numba properly supports the latest Numpy?

hofaflo:
The wheels on conda and PyPI were built with NumPy 1.21, so we don't need a hotfix release, but for the upcoming minor release it would make sense, yes!

cbrnr:
OK - can you push the required changes? Then we should merge this hotfix before any other PRs, because otherwise we don't have any tests.

hofaflo:
Done, thanks @cbrnr!

PR #63, opened 2022-01-10
Warn about joblib only if `n_jobs!=1`

hofaflo:
The warning should not be issued in cases where the user didn't request any parallelism.


cbrnr:
Right, my bad. Thanks @hofaflo!

PR #65, opened 2022-01-10
Make `records_pattern` the first parameter of reader functions

hofaflo:
In most cases, users will not want to pass `data_dir`, as it takes the value from the config file by default. Moving it to the end of the parameter list would allow to pass the record selection without a keyword, so instead of writing this:
```python
mesa = read_mesa(records_pattern='0001')
```
you could write this:
```python
mesa = read_mesa('0001')
```


cbrnr:
Makes sense! Thanks @hofaflo!

PR #55, opened 2021-12-17
Restructure docs

hofaflo:
- While the recursive autosummary we currently use to populate the API reference makes sure we can't forget to add a new function to the docs, it doesn't reflect the actual public API, since it follows the module structure of the package. Listing the public functions and classes manually in `doc/source/api.rst` resolves this issue.
- The doc index page now shows an overview of the doc pages instead of the main readme.
- The main readme's part on heartbeat detection is now a separate page in the docs.

PR #66, opened 2022-01-10
Fix links in heartbeat detection benchmark readme

hofaflo:
Line 2 became outdated when restructuring the docs (#55).
Line 30 has always been wrong.


cbrnr:
Thanks @hofaflo!

PR #68, opened 2022-01-14
Fix `reader_dispatch` not yielding anything

### LINKED ISSUE #67: reader_dispatch doesn't yield any reader
Maybe it's just me, but when I run `python benchmark_detectors.py runtime` in `examples/benchmark`, I do not get any records, i.e. the line `records = list(reader_dispatch(db_slug, data_dir))` yields an empty list. Is this caused by our recent changes in v0.4.0?

hofaflo:
`data_dir` has to be passed as a keyword argument since #65 changed the order of arguments in the reader functions

fixes #67


cbrnr:
Thanks @hofaflo! I think this warrants a hotfix release, could you push v4.0.1 please?

PR #71, opened 2022-01-17
Add option to export records in benchmark

cbrnr:
By default, the config option `export_records` is `False` so records only exported if explicitly requested.

### DIFF
--- a/examples/benchmark/benchmark_detectors.py
+++ b/examples/benchmark/benchmark_detectors.py
@@ -49,13 +49,10 @@ records = list(reader_dispatch(db_slug, data_dir))
 print(f'Loaded {len(records)} records from {db_slug}.')
 
 if cfg.get('export_records', False):
+    from sleepecg import export_record
     print(f'Exporting records to {outfile_dir.resolve()}.')
     for record in records:
-        np.savetxt(
-            outfile_dir / f'{record.id}-{record.lead}.txt',
-            np.atleast_2d(record.ecg),
-            fmt='%.3f',
-        )
+        export_record(record, outfile_dir / f'{record.id}-{record.lead}.txt')
 
 fieldnames = [
     'record_id', 'lead', 'fs', 'num_samples', 'detector', 'max_distance', 'runtime', 'TP',
--- a/examples/benchmark/benchmark_detectors.py
+++ b/examples/benchmark/benchmark_detectors.py
@@ -15,14 +15,18 @@ import yaml
 from tqdm import tqdm
 from utils import detector_dispatch, evaluate_single, reader_dispatch
 
-if len(sys.argv) != 2:
-    print('Usage: python benchmark_detectors.py <benchmark>')
+if len(sys.argv) == 1:
+    print("No benchmark specified, executing 'runtime' benchmark.")
+    benchmark = 'runtime'
+elif len(sys.argv) > 2:
+    print('Usage: python benchmark_detectors.py [<benchmark>]')
     exit()
+else:
+    benchmark = sys.argv[1]
 
 with open('config.yml') as config_file:
     cfg = yaml.safe_load(config_file)
 
-benchmark = sys.argv[1]
 try:
     cfg = cfg[benchmark]
 except KeyError:
@@ -44,6 +48,13 @@ data_dir = Path(cfg.get('data_dir', '~/.sleepecg/datasets')).expanduser()
 records = list(reader_dispatch(db_slug, data_dir))
 print(f'Loaded {len(records)} records from {db_slug}.')
 
+if cfg.get('export_records', False):
+    for record in records:
+        np.savetxt(
+            outfile_dir / f"{record.id}-{record.lead}.txt",
+            np.atleast_2d(record.ecg),
+            fmt="%.3f"
+        )
 
 fieldnames = [
     'record_id', 'lead', 'fs', 'num_samples', 'detector', 'max_distance', 'runtime', 'TP',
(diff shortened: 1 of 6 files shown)


cbrnr:
I checked if `runtime` and `metrics` benchmarks yield identical results when loading the records from the original data files versus from exported text files. It seems like this is the case. Runtimes vary a little of course, but no major changes. Metrics look identical (I just compared the plots). I still have the CSV and SVG result files if you want to take a closer look. Other than that, this is ready for review @hofaflo.

hofaflo:
Yes if you could upload the csv results that would be great!

This exports only the raw ecg data, right? Should we export the annotations as well?

cbrnr:
> Yes if you could upload the csv results that would be great!

I'll send them via email.

> This exports only the raw ecg data, right? Should we export the annotations as well?

Good point. Probably yes, but then we need to create multiple files per record. Is this OK?

hofaflo:
> I'll send them via email.

Looks perfect 🎉

> Good point. Probably yes, but then we need to create multiple files per record. Is this OK?

I guess yes 🤔 Alternatively we could use a csv where the first column is the ecg and the second one a `0` or `1` to indicate R peaks. Of course this wouldn't be the most memory efficient solution, but that's a general issue with the plaintext approach :D We could also include the sampling frequency in some kind of header. 

E.g. this code:
```python
import numpy as np
from sleepecg import read_mitdb

rec = next(read_mitdb())
rpeaks = np.zeros_like(rec.ecg)
rpeaks[rec.annotation] = 1

np.savetxt(
    'test2.csv',
    np.vstack((rec.ecg, rpeaks)).T,
    fmt='%.3f,%.0f',
    header=f'# fs: {rec.fs}Hz\necg,rpeak',
    comments='',
)
```

would produce this:

```
# fs: 360Hz 
ecg,rpeak
-0.145,0
-0.145,0
-0.145,0
-0.145,0
-0.145,0
-0.145,0
-0.145,0
-0.145,0
-0.120,0
-0.135,0
...
```

hofaflo:
Great! We could implement this as a function in io/ecg_readers.py (or as a method of ECGReader, but that would break the functional approach we have so far)?

cbrnr:
I guess we could do both? The implementation should be in a function, and we could optionally add a method which just calls that function.

cbrnr:
Something like that? Note that I used a string for the type annotation because we are going to add a method to `ECGRecord`, so this function needs to come before the class – or does it?

cbrnr:
Or maybe we'll skip the method for now, `ECGRecord` is currently a dataclass so we'd also need to change that. I think the function is sufficient for now.

examples/benchmark/benchmark_detectors.py
```diff
@@ -38,12 +42,17 @@
...
 
 data_dir = Path(cfg.get('data_dir', '~/.sleepecg/datasets')).expanduser()
 records = list(reader_dispatch(db_slug, data_dir))
 print(f'Loaded {len(records)} records from {db_slug}.')
 
+if cfg.get('export_records', False):
+    from sleepecg import export_record
+    print(f'Exporting records to {outfile_dir.resolve()}.')
+    for record in records:
+        export_record(record, outfile_dir / f'{record.id}-{record.lead}.txt')
```
hofaflo, line 55:
```suggestion
        export_record(record, outfile_dir / f'{record.id}-{record.lead}')
```

`export_record` appends `.csv`, so the file would be called e.g. `100-II.txt.csv`

hofaflo:
> Something like that? 

👍 

> Note that I used a string for the type annotation because we are going to add a method to `ECGRecord`, so this function needs to come before the class – or does it?

Functions can be defined below the places they're used, as long as the calling code isn't executed, so both ways would work.

> ECGRecord is currently a dataclass so we'd also need to change that.

Dataclasses can actually have methods, the decorator just creates `__init__`, `__repr__` and comparison dunder methods automatically.

> I think the function is sufficient for now.

I agree, also having only one way is likely less confusing.

Two more notes:
- `export_ecg_record` might be a better name as it's not indended for SleepRecords
- the function should be listed in the docs (i.e. in `api.rst`), either just added to 'Datasets' or we split this category into 'Sleep data' and 'ECG data', WDYT?

examples/benchmark/benchmark_detectors.py
cbrnr: We could also be a bit more strict, i.e. throw an error if a file has an extension other than `.csv`. We could append `.csv` only if there is no extension.

hofaflo: I like that!

cbrnr:
> `export_ecg_record` might be a better name as it's not indended for SleepRecords

Done.

> the function should be listed in the docs (i.e. in `api.rst`), either just added to 'Datasets' or we split this category into 'Sleep data' and 'ECG data', WDYT?

I added it to 'Datasets' for now, splitting would require splitting the menu bar entries in the header as well I assume?

hofaflo:
> I added it to 'Datasets' for now

Seems like that wasn't part of the commit yet :)

> splitting would require splitting the menu bar entries in the header as well I assume?

True, that would be good for consistency, so let's leave it as is.

cbrnr:
I forgot to push the commit. All green now, feel free to merge or let me know if there's anything else.

cbrnr:
Not that it's important, but can/should we link functions in our changelog to the API ref on the docs website? It might be useful unless we have to manually include the corresponding link, in which case it is probably too much effort.

hofaflo:
That would be great, I can't think of a way to have automatic links though :/

cbrnr:
It is possible with .rst but probably not with .md, especially if we want the .md to stay readable e.g. on the GitHub website.

hofaflo:
Hmm... if we keep the chagelog on GitHub I'd prefer to stick with .md, ut if make it part of the docs I think .rst would be fine.

hofaflo:
Anyway this PR is complete, thanks @cbrnr!

PR #76, opened 2022-01-22
Restructure and clarify extras_require

hofaflo:
Currently the optional dependencies are a bit of a mess:
- `full` should be what's needed for the full functionality of the package, but is actually not documented and just used for the CI tests
- `dev` contains `m2r2` which was made obsolete by switching to `myst-parser` in #36
- `dev` should contain everything needed for development (i.e. the full functionality, test and doc requirements), but doesn't

This fixes all of that and documents the different categories :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,7 +95,8 @@ inside your local clone. After that, the checks required in the CI job will be r
 
 
 ## Tests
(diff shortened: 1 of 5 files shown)


cbrnr:
This is awesome! Very nice organization! Thanks @hofaflo!

PR #77, opened 2022-01-22
Require pyedflib v0.1.25 for tests

### LINKED ISSUE #40: Remove warning suppression in `_dummy_mesa_edf`
[This](https://github.com/cbrnr/sleepecg/blob/main/sleepecg/test/test_sleep_readers.py#L26-L28) should be unnecesary once the issue is fixed upstream by holgern/pyedflib#151.

hofaflo:
resolves #40 

The underlying issue was fixed in v0.1.24 but that version required NumPy>=1.22 (see holgern/pyedflib#164), so we should require v0.1.25 instead.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,7 +52,7 @@ dev =
     numpydoc>=1.1.0
     pandas>=1.2.0
     pre-commit>=2.13.0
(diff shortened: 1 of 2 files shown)


cbrnr:
Thanks @hofaflo!

PR #78, opened 2022-01-25
Add classification

hofaflo:
Sorry for the massive PR 😅


doc/source/classification.md
```diff
@@ -0,0 +1,49 @@
+(classification)=
+# Classification
+
+
+## Included classifiers
+SleepECG contains classifiers trained on large open sleep datasets.
+Note that the performance metrics are averages, so a classifier might perform much better or much worse for an individual record.
+
+|Identifier|Stages|Model|Training set (nights)|Test set (nights)|Acc.|κ|Source|
```
cbrnr, line 9:
```suggestion
|Identifier|Stages|Model|Training set (nights)|Test set (nights)|Acc|κ|Source|
```

doc/source/classification.md
```diff
@@ -0,0 +1,49 @@
...
+## Contributing
+You are very welcome to add new classifiers to SleepECG!
+To do so, please submit a [pull request](https://github.com/cbrnr/sleepecg/pulls).
+Make sure it contains the following elements:
+- The script required to replicate and validate the classifer, placed in `examples/classifiers`.
+- The classifier itself, saved as a `.zip` file with `save_classifier`, placed in `sleepecg/classifiers`.
+- An entry in the table above.
+
+
+## Sleep stage labels & `stages_mode`
```
cbrnr, line 29:
```suggestion
## Sleep stage labels and `stages_mode`
```

doc/source/classification.md
```diff
@@ -0,0 +1,49 @@
...
+|-|:-:|:-:|:-:|:-:|:-:|:-:|
+|`'wake-sleep'`         |UNDEFINED|SLEEP|WAKE |    |    |    |
+|`'wake-rem-nrem'`      |UNDEFINED|NREM |REM  |WAKE|    |    |
+|`'wake-rem-light-n3'`  |UNDEFINED|N3   |LIGHT|REM |WAKE|    |
+|`'wake-rem-n1-n2-n3'`  |UNDEFINED|N3   |N2   |N1  |REM |WAKE|
+
+
+The decision against keeping the integer values for `REM` and `WAKE` constant between modes was made because labels starting at 0 and without gaps simplify working with one-hot encoded stages (or probabilities) and avoid issues when calculating sample weights.
+
+When working with SleepECG you usually will not need to worry about this, just remember to pass the correct `stages_mode` to the functions expecting the parameter :)
```
cbrnr, line 49:
```suggestion
When working with SleepECG, you usually will not need to worry about this, just remember to pass the correct `stages_mode` to the functions expecting the parameter.
```

doc/source/configuration.md
```diff
@@ -8,3 +8,4 @@ This table lists the possible configuration settings and where they are used:
 |Key|Default value|Description|Used in|
 |-|-|-|-|
 |`data_dir`|`'~/.sleepecg/datasets'`|Used as the default location to store the files downloaded by all reader functions. Data downloaded for tests is also stored there.|`read_ltdb`, `read_mitdb`, `read_gudb`, `read_mesa`, `read_shhs`, `read_slpdb`|
+|`classifiers_dir`|`'~/.sleepecg/classifiers'`|Used as the default location to save and load `SleepClassifer` objects..|`list_classifiers`, `load_classifier`, `save_classifier`|
```
cbrnr, line 11:
```suggestion
|`classifiers_dir`|`'~/.sleepecg/classifiers'`|Used as the default location to save and load `SleepClassifer` objects.|`list_classifiers`, `load_classifier`, `save_classifier`|
```

examples/classifiers/ws_gru_mesa.py
```diff
@@ -0,0 +1,118 @@
...
+    feature_selection=[
+        'hrv-time',
+        'hrv-frequency',
+        'recording_start_time',
+        'age',
+        'gender',
+    ],
+    min_rri=0.3,
+    max_rri=2,
+    max_nans=0.5,
```
cbrnr, line 32: Maybe put that in a dict because it is used in `save_classifier` as well?

examples/classifiers/ws_gru_mesa.py
```diff
@@ -0,0 +1,118 @@
...
+model.fit(
+    features_train_pad,
+    stages_train_pad,
+    epochs=5,
+    sample_weight=sample_weight,
+)
+
+# %% Store classifier
+save_classifier(
+    name='ws-gru-mesa',
```
cbrnr, line 77: I think we should explain the name. `ws` is wake/sleep, `mesa` is also clear, but what does `gru` mean? This should also mentioned in the docs.

setup.cfg
```diff
@@ -56,6 +59,7 @@ dev =
     pytest>=6.2.0
     setuptools>=56.0.0
     sphinx>=4.1.0
+    tensorflow>=2.7.0
```
cbrnr, line 62: I don't get the difference between `tensorflow` and `keras`. We're using `keras`, so why is the `tensorflow` package required? There is also a `keras` package on PyPI...

sleepecg/classification.py
```diff
@@ -0,0 +1,756 @@
...
+
+
+def prepare_data_keras(
+    features: List[np.ndarray],
+    stages: List[np.ndarray],
+    stages_mode: str,
+    mask_value: int = -1,
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
+    """
+    Mask+pad data and calculate sample weights for use with a keras model.
```
cbrnr, line 105:
```suggestion
    Mask and pad data and calculate sample weights for a keras model.
```

sleepecg/classification.py
```diff
@@ -0,0 +1,756 @@
...
+
+    The following steps are performed:
+
+    - Merge sleep stages in `stages` according to `stage_mode`.
+    - Set features corresponding to `SleepStage.UNDEFINED` to `mask_value`.
+    - Replace `np.nan` and `np.inf` in `features` with `mask_value`
```
cbrnr, line 111:
```suggestion
    - Replace `np.nan` and `np.inf` in `features` with `mask_value`.
```

sleepecg/classification.py
```diff
@@ -0,0 +1,756 @@
...
+    feature_extraction_params : dict[str, typing.Any]
+        The parameters passed to :func:`extract_features`, as a dictionary
+        mapping string parameter names to values. Should not include
+        `records` and `n_jobs`.
+    mask_value : int, optional
+        Only required for keras models, as passed to `prepare_data_keras`
+        and `keras.layers.Masking`, by default `None`.
+    classifiers_dir : str | pathlib.Path, optional
+        Directory in which the `.zip` file is stored. If `None`
+        (default), the value is be taken from the configuration.
```
cbrnr, line 236:
```suggestion
        (default), the value is taken from the configuration.
```

sleepecg/classification.py
```diff
@@ -0,0 +1,756 @@
...
+
+    def __str__(self) -> str:
+        """
+        Create a string representation of a `SleepClassifier`.
+
+        Returns
+        -------
+        str
+            The string representation.
+        """
```
cbrnr, line 311: I don't think we need a docstring for this standard method.

sleepecg/classification.py
```diff
@@ -0,0 +1,756 @@
...
+    """
+
+    model: Any
+    stages_mode: str
+    feature_extraction_params: Dict[str, Any]
+    model_type: str
+    mask_value: Optional[int] = None
+    source_file: Optional[str] = None
+
+    def __str__(self) -> str:
```
cbrnr, line 303: Do we also want/need a `__repr__`?

sleepecg/classification.py
```diff
@@ -0,0 +1,756 @@
...
+    mask_value : int, optional
+        Only required for keras models, as passed to `prepare_data_keras`
+        and `keras.layers.Masking`, by default `None`.
+    classifiers_dir : str | pathlib.Path, optional
+        Directory in which the `.zip` file is stored. If `None`
+        (default), the value is be taken from the configuration.
+
+    See Also
+    --------
+    load_classifier : Load classifiers saved with this function
```
cbrnr, line 240:
```suggestion
    load_classifier : Load classifiers saved with this function.
```

sleepecg/classification.py
```diff
@@ -0,0 +1,756 @@
...
+        Confusion matrix whose i-th row and j-th column entry indicates the
+        number of samples with true label being i-th class and predicted
+        label being j-th class.
+    """
+    return np.bincount(N * y_true + y_pred, minlength=N*N).reshape(N, N)
+
+
+def _cohen_kappa(confmat: np.ndarray) -> float:
+    """
+    Cohen's kappa: a statistic that measures inter-annotator agreement.
```
cbrnr, line 445:
```suggestion
    Compute Cohen's kappa (which measures inter-annotator agreement).
```

sleepecg/classification.py
```diff
@@ -0,0 +1,756 @@
...
+def evaluate(
+    stages_true: np.ndarray,
+    stages_pred: np.ndarray,
+    stages_mode: str,
+    show_undefined: bool = False,
+) -> None:
+    """
+    Evaluate the performance of a sleep stage classifier.
+
+    Prints overall accuracy, cohen's kappa, confusion matrix and per-class
```
cbrnr, line 517:
```suggestion
    Prints overall accuracy, Cohen's kappa, confusion matrix and per-class
```

cbrnr (commented):
This is really great! I added some minor comments, and I am now going to try the example to make sure it works on my computer. But you can already go ahead and address my comments in the meantime.

examples/classifiers/ws_gru_mesa.py
hofaflo: The table on the [doc page](https://sleepecg--78.org.readthedocs.build/en/78/classification.html#included-classifiers) contains a link to the wikipedia page of [gated recurrent units].

We could add a comment like this?
```
# Classifier name explanation:
# ws -> WAKE-SLEEP
# gru -> main component of the classifier network, a gated recurrent unit
# mesa -> MESA dataset (https://sleepdata.org/datasets/mesa)
```

setup.cfg
hofaflo: `keras` is the high-level interface that provides 'layers' which make it easy to build networks. `tensorflow` is the backend that implements all the math and gradient computation stuff. So we actually need both, but `keras` comes as a part of `tensorflow`, therefore the dependency is written like that.

sleepecg/classification.py
hofaflo: Me neither, but the numpydoc setting for pydocstyle says otherwise 🤷 We could switch that off alternatively?

sleepecg/classification.py
hofaflo: I'd prefer not to, as it would be difficult to follow what the [Python docs](https://docs.python.org/3/library/functions.html#repr) here.

examples/classifiers/ws_gru_mesa.py
cbrnr: I didn't see this link, I think it's fine then, no need to explain further.

sleepecg/classification.py
cbrnr: How?

sleepecg/classification.py
cbrnr: Yes, but it could at least be more informative than the default memory address repr. For interactive use it's quite handy to just type the name of the object instead of an explicit print.

sleepecg/classification.py
hofaflo: By adding `add_ignore = D105` to the pydocstyle config in `setup.cfg`. `D105` refers to "Missing docstring in magic method" ([link](http://www.pydocstyle.org/en/stable/error_codes.html)).

cbrnr: Seems worth it, WDYT? We won't ever need this I think.

hofaflo: Likely not, I'll make the changes!

sleepecg/classification.py
cbrnr: But would a more conventional one-line `__repr__` also work, e.g.

```
f'SleepClassifier(model, {self.stages_mode.upper()}, feature_extraction_params, model_type={self.model_type}'
```

etc.?

What does the default `repr` look like? Do you have an example?

hofaflo: that's what the default repr looks like:
```
SleepClassifier(model=<keras.engine.sequential.Sequential object at 0x000001DF1CF33108>, stages_mode='wake-sleep', feature_extraction_params={'feature_selection': ['hrv-time', 'hrv-frequency', 'recording_start_time', 'age', 'gender'], 'lookback': 240, 'lookforward': 270, 'max_nans': 0.5, 'max_rri': 2, 'min_rri': 0.3}, model_type='keras', mask_value=-1, source_file='c:\\users\\florian\\desktop\\masterarbeit\\code\\sleepecg\\sleepecg\\classifiers\\ws-gru-mesa.zip')
```

I didn't realize it contains that much information (thanks to `SleepClassifier` being a dataclass), so I'd actually vote for keeping the `__str__` as is, and leaving the default `__repr__`.

cbrnr: Yes, that's probably the best solution for now. I thought about making the repr a bit more concise, but there's no way it could contain all of this information. In MNE, we'd ignore the part that the repr should be a string to create the object. Instead, we only show the most important bits of information, so this could be:

```
<SleepClassifier | keras, wake-sleep, ws-gru-mesa.zip>
```

And `__str__` could then show more (all?) information. WDYT?

hofaflo: > ```
> <SleepClassifier | keras, wake-sleep, ws-gru-mesa.zip>
> ```

I like that! So to summarize, we keep the current implementation of `__str__`, and add a `__repr__` in the style above?

hofaflo: I added the `__repr__` in 8e93dd6, with a two modifications:
- `SleepClassifier.source_file` is now a `pathlib.Path`, to make getting the name cleaner
- `stages_mode` comes before `model_type`, as I think it's more important

sleepecg/classification.py
```diff
@@ -0,0 +1,753 @@
...
+        stages_pred_probs = stages_pred
+        stages_pred = stages_pred_probs.argmax(1)
+    if record.sleep_stages is not None:
+        num_subplots += 1
+    if show_bpm:
+        num_subplots += 1
+
+    start_time = _time_to_sec(record.recording_start_time)
+
+    fig, ax = plt.subplots(num_subplots, sharex=True, figsize=(7, 4))
```
cbrnr, line 688:
```suggestion
    fig, ax = plt.subplots(num_subplots, sharex=True, figsize=(10, 4))
```
```suggestion
    fig, ax = plt.subplots(num_subplots, sharex=True, figsize=(7, 4))
```

cbrnr:
When running `try_ws_gru_mesa.py`, I get a buch of these warnings:

```
2022-02-03 07:23:10.299351: W tensorflow/core/common_runtime/graph_constructor.cc:803] Node 'cond/while' has 14 outputs but the _output_shapes attribute specifies shapes for 48 outputs. Output shapes may be inaccurate.
```

Is this expected? I also get these warnings:

```
/Users/clemens/Repositories/sleepecg/sleepecg/feature_extraction.py:321: RuntimeWarning: HR analysis window too short for estimating PSD for feature VLF. 3030.3s required, got 510s
  warnings.warn(msg, category=RuntimeWarning)
/Users/clemens/Repositories/sleepecg/sleepecg/feature_extraction.py:221: RuntimeWarning: divide by zero encountered in true_divide
  cvSD = SDSD / np.nanmean(SD, axis=1)
```

Should we silence these warnings in the example? Or is it better not to?

hofaflo:
ad tensorflow warnings: apparently these arise because the model was created using tensorflow 2.7 and you're probably using 2.8? I ran the evaluation again using 2.8 and nothing changed, so maybe we just add a comment in `try_ws_gru_mesa.py` telling users they don't need to worry about that warning?

ad feature_extraction warnings: I'd rather not silence them in the example to keep the code simple there. Maybe instead in `stage`, since they're mainly there to inform users who create a classifier, but in case that's already been done and the evaluation showed it works nonetheless, end users won't care?

cbrnr:
OK, a quick note on the Tensorflow warnings would be useful (maybe state that the model was created with TF 2.7, and that running with a higher version should in general work but might create some warnings or something like that).

And yes, keeping the other warnings in the example is fine.

cbrnr:
Nice, thanks @hofaflo!

PR #82, opened 2022-03-13
Use lighter font weight for doc headings

hofaflo:


cbrnr:
Internal links are blue and bold - should we also at least make them not bold (so they look like external links)?

hofaflo:
Fine with me, I have no preference there.

cbrnr:
Thanks @hofaflo!

PR #83, opened 2022-03-17
Improve doc footer

hofaflo:
A few small changes:
- remove the year in the copyright notice to avoid updating it
- add a "last updated on" line to let users know the project is active
- remove the "author" variable in conf.py (this doesn't seem to be used anywhere)

@cbrnr, we talked about removing the word "Copyright", but since this requires a custom template, I'd like to keep it as-is.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -12,7 +12,6 @@ import sleepecg
 
 # -- Project information --------------------------------------------------
 project = 'SleepECG'
-author = 'Florian Hofer'
 copyright = 'SleepECG Developers'
 version = sleepecg.__version__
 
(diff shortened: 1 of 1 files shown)


cbrnr:
Looks good, but don't you think three lines in the footer is a bit much? I kind of like how we did it for MNELAB where we put the Sphinx attribution in a separate page instead of the footer.

hofaflo:
I think it's fine like that, on most pages it's only visible when scrolling to the bottom anyway. I actually prefer this over an additional page, but if you have a strong preference we can change it too :)

cbrnr:
If I really don't like it we can always change it later. Thanks @hofaflo!

PR #90, opened 2022-05-20
Improve docstring for heartbeats function

raphaelvallat:
Hi @cbrnr and @hofaflo,

I have just added some examples in the docstring of the functions. I think this will be useful. Feel free to edit the text.

Thanks,
Raphael

### DIFF
--- a/doc/source/heartbeat_detection.md
+++ b/doc/source/heartbeat_detection.md
@@ -24,7 +24,7 @@ from sleepecg import detect_heartbeats
 ecg = electrocardiogram()  # 5 min of ECG data at 360 Hz
 fs = 360
 beats = detect_heartbeats(ecg, fs)
-print(f"{len(beats)} heartbeats detected ({", ".join(beats[0:5])}, ...)")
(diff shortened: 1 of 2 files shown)


cbrnr:
Thanks @raphaelvallat, these examples will be very helpful indeed. However, since they are already pretty long, what do you think about including them in the documentation instead of in the docstrings? We could create an "Example" section e.g. right after "Usage" on https://sleepecg.readthedocs.io/en/stable/heartbeat_detection.html?

If you think examples would be useful in docstrings, I'd prefer shorter examples (that we could maybe also doctest automatically).

raphaelvallat:
Hi @cbrnr,

I tend to only read the API and not really the documentation, so for me having at least a minimal example in the docstring is definitely useful. But we can shorten what I have done, and include the full example in the heartbeat_detection.md file.

Also, I do think that the doctest would work even with the current (long) examples.

raphaelvallat:
@cbrnr done. Feel free to edit!

cbrnr:
Thanks @raphaelvallat!

PR #91, opened 2022-05-20
Improvements in preprocess_rri

raphaelvallat:
- [x] Previously, preprocess_rri was applied in-place, which can be confusing and error-prone. We now return a copy of the array, leaving the original array unchanged.
- [x] Previously, the function would fail if a list was used as the input. We now force the conversion to a numpy.array of dtype float.
- [x] Added an example in docstring

Thanks,
Raphael

### DIFF
--- a/sleepecg/feature_extraction.py
+++ b/sleepecg/feature_extraction.py
@@ -471,13 +471,20 @@ def preprocess_rri(
     -------
     np.ndarray
         The cleaned RRI series.
+
+    Examples
+    --------
+    Mask RR intervals outside the range of 0.4 to 2 seconds (= 30 to 150 bpm)
+
+    >>> from sleepecg import preprocess_rri
+    >>> preprocess_rri([0.5, 0.2, 0.8, 2.5, 0.6], min_rri=0.4, max_rri=2)  # noqa
+    array([0.5, nan, 0.8, nan, 0.6])
     """
-    invalid_rri = np.zeros_like(rri, dtype=bool)
+    rri = np.asarray(rri, dtype=float)
     if min_rri is not None:
-        invalid_rri |= rri < min_rri
+        rri[rri < min_rri] = np.nan
     if max_rri is not None:
-        invalid_rri |= rri > max_rri
-    rri[invalid_rri] = np.nan
+        rri[rri > max_rri] = np.nan
     return rri
 
 
--- a/sleepecg/feature_extraction.py
+++ b/sleepecg/feature_extraction.py
@@ -474,10 +474,11 @@ def preprocess_rri(
 
     Examples
     --------
-    Mask RR intervals outside the range of 0.4 to 2 seconds (= 30 to 150 bpm)
+    Mask RR intervals outside the range of 0.4 to 2 seconds
+    (= 30 to 150 bpm)
 
     >>> from sleepecg import preprocess_rri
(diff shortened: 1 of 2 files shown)


sleepecg/feature_extraction.py
```diff
@@ -471,13 +471,20 @@ def preprocess_rri(
     -------
     np.ndarray
         The cleaned RRI series.
+
+    Examples
+    --------
+    Mask RR intervals outside the range of 0.4 to 2 seconds (= 30 to 150 bpm)
+
+    >>> from sleepecg import preprocess_rri
+    >>> preprocess_rri([0.5, 0.2, 0.8, 2.5, 0.6], min_rri=0.4, max_rri=2)  # noqa
```
hofaflo, line 480: why do we need the `# noqa` here?

sleepecg/feature_extraction.py
```diff
@@ -471,13 +471,20 @@ def preprocess_rri(
...
+    array([0.5, nan, 0.8, nan, 0.6])
     """
-    invalid_rri = np.zeros_like(rri, dtype=bool)
+    rri = np.asarray(rri, dtype=float)
```
hofaflo, line 483:
```suggestion
    rri = np.array(rri, dtype=float)
```
[`np.asarray`](https://numpy.org/doc/stable/reference/generated/numpy.asarray.html) doesn't perform a copy if the input matches the desired dtype and order, so I think [`np.array`](https://numpy.org/doc/stable/reference/generated/numpy.array.html) would be the safest option here

hofaflo (changes requested):
Nice, thanks for catching the in-place bug!

sleepecg/feature_extraction.py
raphaelvallat: To avoid flake8 warning:

image

But feel free to remove!

hofaflo: Ah, I see, that should be the "Mask RR intervals [...]" line, right? So I'd remove this here and shorten the other line a bit :)

```suggestion
    >>> preprocess_rri([0.5, 0.2, 0.8, 2.5, 0.6], min_rri=0.4, max_rri=2)
```

sleepecg/feature_extraction.py
```diff
@@ -471,13 +471,20 @@ def preprocess_rri(
     -------
     np.ndarray
         The cleaned RRI series.
+
+    Examples
+    --------
+    Mask RR intervals outside the range of 0.4 to 2 seconds (= 30 to 150 bpm)
```
hofaflo, line 477:
```suggestion
    Mask RR intervals outside the range of 0.4 to 2 seconds (30 to 150 bpm)
```

sleepecg/feature_extraction.py
```diff
@@ -471,13 +471,20 @@ def preprocess_rri(
...
+
+    >>> from sleepecg import preprocess_rri
+    >>> preprocess_rri([0.5, 0.2, 0.8, 2.5, 0.6], min_rri=0.4, max_rri=2)  # noqa
+    array([0.5, nan, 0.8, nan, 0.6])
     """
-    invalid_rri = np.zeros_like(rri, dtype=bool)
+    rri = np.array(rri, dtype=float)
```
cbrnr, line 483:
```suggestion
    rri = np.array(rri, dtype=float)  # make a copy
```

cbrnr:
Please also make sure to add a changelog entry, because returning a copy is a big change in behavior!

raphaelvallat:
Done!

Issue #87, opened 2022-05-02
The function detect_heartbeats is not handling the absence of signal correctly


LegrandNico:
The function `detect_heartbeats()` can be easily confused when presented with time series containing a flat signal e.g. at the beginning of the recording.

```python
from scipy.misc import electrocardiogram
from sleepecg import detect_heartbeats

# This works as expected
ecg = electrocardiogram()
detect_heartbeats(ecg, fs=1000)

# However, even when presented with 60 seconds of flat signal, the function still returns (a lot of) peaks
detect_heartbeats(np.repeat(-1.5, 60000), fs=1000)

```

The problem is that the flat signal at the beginning or at the end of the time series is not discarded automatically. Of course, better performance is achieved by providing better quality data, but I think this is something that should be automatically handled by the function: no peak detection on the portion of the signal consisting of consecutive constant values over at least n samples, at the beginning or at the end (ideally also if this happened during the recording).

In addition to that, the erroneous detection of (lot of) peaks in these segments appears to considerably slow down the function itself, the following time series (~1h ECG recording starting with ~45 s of flat signal) can for example completely crash the function that will hang forever (or at least for a very long time).
[ecg.zip](https://github.com/cbrnr/sleepecg/files/8602442/ecg.zip)

cbrnr:
Yes, this is a property of the Pan-Tompkins detector, which expects a heartbeat after a certain amount of time. I agree we should have some kind of (optional) preprocessing, which detects e.g. flat segments (which is relatively easy) and extremely noisy segments (which is probably hard). In such segments, no heartbeats should be detected. The thresholds of the detector should probably remain constant until the next good data segment – but this is also open for discussion.

I think one way to detect flat segments could be to only perform a certain number of threshold decreases. I think currently we do it 16x, and then we just take the peak that's in that signal. We could instead say that if there is no clear peak after e.g. 8x lowering the thresholds, we skip it. But this wouldn't be the classic Pan-Tompkins anymore, so I'd prefer a separate preprocessing step.

How do other packages solve this issue?

LegrandNico:
Yes, I think the classic Pan-Tompkins version implemented here should at least be fixed so that it does not run forever with signals like the one I shared, so skipping after a given number of thresholds could maybe solve this. 

[...]

PR #92, opened 2022-06-06
Handle flat data at the beginning of the recording

raphaelvallat:
Fixes https://github.com/cbrnr/sleepecg/issues/87

Flat data at the beginning of the ECG recording can mess up the Pan & Tompkins heartbeat detection. This PR adds a fix for that, such that the heartbeat detection only starts after the first non-flat sample of the data. The index of the first non flat sample is then added back to the array of heartbeat indices.

The overhead of calculating `np.nonzero(np.diff(ecg))[0]` is ~100 ms for a 8 hour ECG signal sampled at 256 Hz, on a 2019 MacBook Pro:

```python
%timeit np.nonzero(np.diff(np.random.rand(256 * 8 * 3600)))
```

I have also added unit tests and updated the changelog.

Feel free to edit,

Thanks,
Raphael

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 ## [0.5.1] - 2022-07-21
 ### Added
 - Add more tests and documentation for the Pan-Tompkins detector ([#89](https://github.com/cbrnr/sleepecg/pull/89) by [Raphael Vallat](https://github.com/raphaelvallat))
-- Better handling of flat data for the Pan-Tompkins detector ([#87](https://github.com/cbrnr/sleepecg/pull/87) by [Raphael Vallat](https://github.com/raphaelvallat))
+- The Pan-Tompkins detector detects flat data at the beginning of a recording to avoid messing up its thresholds ([#87](https://github.com/cbrnr/sleepecg/pull/87) by [Raphael Vallat](https://github.com/raphaelvallat))
 
 ### Changed
 - The `preprocess_rri` function no longer operates in-place and instead returns a copy of the input array ([#91](https://github.com/cbrnr/sleepecg/pull/91) by [Raphael Vallat](https://github.com/raphaelvallat))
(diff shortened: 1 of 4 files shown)


cbrnr:
There are some style issues. Also, 100ms is not much in absolute time, but it almost doubles the detection time for the C and Numba backends. I'd rather not do this by default, but maybe introduce a parameter (which by default does *not* detect flat segments). Not sure what a good name for this parameter would be though.

cbrnr:
I like the explicit condition better.

raphaelvallat:
Done — I have added a unit test to catch the ValueError. Let me know if you need anything else from me. Thanks

cbrnr:
Thanks! If possible, I'd add a more specific test for the `ValueError` (so include at least part of the message). Other than that, this looks great! I will do my benchmarks to see how this performs, and then we can decide on having a new parameter or not.

cbrnr:
@LegrandNico could you test if this PR fixes the issues you had with one of your data sets?

cbrnr:
@raphaelvallat currently we set all samples before the first zero-crossing to 0:

https://github.com/cbrnr/sleepecg/blob/main/sleepecg/heartbeats.py#L128-L131

I wonder (1) if we can use this step also for detecting flat segments and (2) why setting 2s of flat data is OK (better?) than longer flat segments.

@hofaflo maybe you have some insights?

LegrandNico:
> @LegrandNico could you test if this PR fixes the issues you had with one of your data sets?

Yes the function is working correctly now 👍

raphaelvallat:
Hi here, is there any plan to merge and release this PR? Thanks!

cbrnr:
Sorry that it's taking so long. Yes, I will merge this PR soon (today?). A new release is also planned sometime soon, but probably not within the next 3-4 weeks since I will be on vacation then.

cbrnr:
Thanks @raphaelvallat!

PR #94, opened 2022-07-20
Bump dependencies

cbrnr:
Python 3.8 is now the oldest supported version.


hofaflo:
ea29bcd removes the link to the GitHub repo. Did you want to include it somewhere else or to actually remove it completely?

cbrnr:
I've restored the icons in the footer, one of which links to the GitHub repo. I thought this should be enough - do you think it's not easy to discover?

hofaflo:
Ah, that's perfect! I must have overlooked that, sorry for the false alarm.

PR #99, opened 2022-07-22
Try customizing Furo without a template

cbrnr:
I think "Made with..." and "Prev"/"Next" should be removable with settings in `conf.py`.


cbrnr:
Seems like this is not possible with settings.

Issue #95, opened 2022-07-20
Minor documentation website issues


cbrnr:
- [ ] All tables are centered in the latest version, but they should be left-aligned.
- [ ] Some function names are too long to fit into the sidebar API dropdown menu. Maybe there is a way to avoid this?

PR #100, opened 2022-07-22
Tweak docs

cbrnr:
Add parentheses after functions, do not use the linkcode extension (Sphinx should be able to detect all module functions/classes automatically).

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -53,6 +53,7 @@ autodoc_mock_imports = ["scipy", "tqdm"]
 autodoc_typehints = "none"
 autosummary_generate = True
 html_show_sourcelink = False
+add_function_parentheses = True
 
 numpydoc_class_members_toctree = False
 numpydoc_show_class_members = False
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -42,6 +42,8 @@ myst_substitutions = {
     "version": version,
 }
 
+myst_heading_anchors = 3
+
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3", None),
     "numpy": ("https://numpy.org/devdocs", None),
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -4,7 +4,7 @@
 # full list see the documentation:
 # https://www.sphinx-doc.org/en/master/usage/configuration.html
 
-# import inspect
+import inspect
 import os
 import sys
 
@@ -23,7 +23,7 @@ extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
     "sphinx.ext.intersphinx",
-    # "sphinx.ext.linkcode",
+    "sphinx.ext.linkcode",
     "numpydoc",
     "myst_parser",
 ]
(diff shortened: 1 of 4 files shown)


cbrnr:
Issues so far:

- [ ] Cross referencing functions does not work, see e.g. https://sleepecg--100.org.readthedocs.build/en/100/heartbeat_detection.html#usage
- [ ] Functions in the API section are only linked if they are in a `:toctree:` – but I don't want them to show up in the left sidebar...

cbrnr:
None, I think API itself is sufficient. This would also solve the issue that longer function names don't fit in a row.

hofaflo:
Cross-referencing needs a dot before the function name for some reason (alternatively, the fully qualified name, e.g. sleepecg.detect_heartbeats` works too).

Hiding the sidebar entries can be done by placing `:tocdepth: 1` at the top of `api.rst` (with an empty line before the page title). I like being able to explore a package's functions through the sidebar, but if you prefer having just the "API" entry it's fine for me as well.

cbrnr:
> Cross-referencing needs a dot before the function name for some reason (alternatively, the fully qualified name, e.g. sleepecg.detect_heartbeats` works too).

OK, let's use the long form then.

> Hiding the sidebar entries can be done by placing :tocdepth: 1 at the top of api.rst (with an empty line before the page title). I like being able to explore a package's functions through the sidebar, but if you prefer having just the "API" entry it's fine for me as well.

I'd prefer the latter, the API page has an overview of all functions anyway. It makes the website less cluttered I think.

> Linkcode isn't required to detect functions and classes, but to provide links to the source code on GitHub. I'd very much like to keep that.

Yes, I didn't know that. I'll restore it.

cbrnr:
Alright, now the only thing remaining are the centered tables, which I'd like to be left-aligned (like they used to). Not sure what caused that change, but if it's not something that I did we should probably report it upstream. Any ideas?

hofaflo:
When was the latest point in time that you remember them being left-aligned?

A workaround would be to add this to  `_static/custom.css`:

```
table {
    margin-left: 0 !important;
}
```

Or, what I would like even better since we only have rather wide tables anyway, make them take up the entire width:

```
table {
    width: 100%;
}
```

doc/source/conf.py
```diff
@@ -10,12 +10,15 @@
 
 import sleepecg
 
-# -- Project information --------------------------------------------------
+# -- Path setup ----------------------------------------------------------------------------
+sys.path.insert(0, os.path.abspath("../../sleepecg"))
```
hofaflo, line 14: Why is this necessary?

cbrnr: Seems like it isn't. I read that [here](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/sphinx-config.html#autodoc-configuration), so I went ahead and added it. Do you know why it is not required?

cbrnr:
> When was the latest point in time that you remember them being left-aligned?

Before I started messing with everything 😄, so possibly before I bumped our dependencies. Is there a way we can access older versions of our docs?

hofaflo:
> Is there a way we can access older versions of our docs?

Not that I know of, but we can pin the furo version to find which release changed this.

cbrnr:
I think it looks good now with the custom CSS. Please go ahead and merge if you are happy. It's probably not worth hunting down the change, unless it is this one: https://github.com/pradyunsg/furo/commit/08ba46d070bfb9e0e013586822e073742dbb372c

doc/source/conf.py
hofaflo: Before building the docs locally or on RTD, the package is installed, so it's available in the pythonpath anyway.

hofaflo:
> unless it is this one: [pradyunsg/furo@08ba46d](https://github.com/pradyunsg/furo/commit/08ba46d070bfb9e0e013586822e073742dbb372c)

Judging from the commit's date (2021-10-09), I'd say this is unlikely. Also I think this only changes the alignment of the table content, not the entire table.

hofaflo:
Closes #95, thanks @cbrnr!

PR #105, opened 2022-08-01
Fix type errors

### LINKED ISSUE #93: Fix type annotation issues
I turned on basic type checking for the first time (didn't know that you had to opt in), and boy there are a lot of errors. Since we are using type annotations, we should definitely fix all the errors (otherwise why are we using annotations in the first place).

cbrnr:
Fixes #93.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,3 +24,14 @@ archs = 'x86_64 arm64'
 
 [tool.black]
 line-length = 92
+
+
+[tool.mypy]
+packages = ['sleepecg']
+exclude = '^sleepecg/test/'
+ignore_missing_imports = true
+disallow_untyped_defs = true
+no_implicit_optional = true
+warn_unreachable = true
+strict_equality = true
+pretty = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,6 @@ packages = ['sleepecg']
 exclude = '^sleepecg/test/'
 ignore_missing_imports = true
 disallow_untyped_defs = true
-no_implicit_optional = true
 warn_unreachable = true
 strict_equality = true
 pretty = true
--- a/sleepecg/io/gudb.py
+++ b/sleepecg/io/gudb.py
@@ -5,7 +5,7 @@
 """Utilities for downloading GUDB data."""
 
 from pathlib import Path
-from typing import Optional, Union
+from typing import Dict, Optional, Union
 
 from ..config import get_config
 from .utils import _calculate_checksum
@@ -368,7 +368,7 @@ GUDB_MD5 = {
 }
 
 
-def _generate_gudb_md5(data_dir: Optional[Union[str, Path]] = None):
+def _generate_gudb_md5(data_dir: Optional[Union[str, Path]] = None) -> Dict[str, str]:
     """
     Compute checksums for files in GUDB.
 
--- a/sleepecg/io/gudb.py
+++ b/sleepecg/io/gudb.py
@@ -4,8 +4,10 @@
 
 """Utilities for downloading GUDB data."""
 
+from __future__ import annotations
+
 from pathlib import Path
-from typing import Dict, Optional, Union
+from typing import Optional
 
 from ..config import get_config
 from .utils import _calculate_checksum
@@ -368,7 +370,7 @@ GUDB_MD5 = {
 }
 
 
-def _generate_gudb_md5(data_dir: Optional[Union[str, Path]] = None) -> Dict[str, str]:
+def _generate_gudb_md5(data_dir: Optional[str | Path] = None) -> dict[str, str]:
     """
     Compute checksums for files in GUDB.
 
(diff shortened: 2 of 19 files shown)


cbrnr:
In my opinion, the remaining errors/warnings (when setting `"python.analysis.typeCheckingMode": "basic"`) are useless. These include:

- values that can be `None`, even if we explicitly take care of that literally in the previous line
  ```python
      if classifiers_dir is None:
        classifiers_dir = get_config("classifiers_dir")

    target_file = Path(classifiers_dir).expanduser() / name
  ```
- things like `ax[row]`, where Pylance says that we cannot index a Matplotlib Axes,
- and NumPy-related stuff where I cannot even read the error message because it is basically gibberish.

@hofaflo let me know if you think we should address any of the remaining errors (and if so, please show me how).

cbrnr:
Ima turn off type checking now to stay sane.

hofaflo:
Pyright will still complain, but I think making mypy happy is enough :)

The mkdocstrings issue is already fixed upstream (mkdocstrings/python#45), so I bumped the dependency. 5e81986 should have worked around this, but for some reason readthedocs ignored the changed path to the configuration file and kept calling mkdocs with `--config-file mkdocs.yml` instead of `--config-file ./mkdocs.yml` (see the [build log](https://readthedocs.org/projects/sleepecg/builds/18683094/)) :shrug:

hofaflo:
This is ready to merge from my point of view @cbrnr :)

sleepecg/io/physionet.py
```diff
@@ -21,8 +21,8 @@
 def _list_physionet(
     data_dir: Path,
     db_slug: str,
-    db_version: Optional[str] = "1.0.0",
-    pattern="*",
+    db_version: str = "1.0.0",
```
cbrnr, line 24: I see that you enabled an option to not require `Optional`, but why is `Optional` still used in other places (like e.g. `_generate_gudb_md5()`)?

sleepecg/io/gudb.py
```diff
@@ -5,7 +5,7 @@
 """Utilities for downloading GUDB data."""
 
 from pathlib import Path
-from typing import Optional, Union
+from typing import Dict, Optional, Union
```
cbrnr, line 8: Could we use the normal `dict` as a type hint? I know that this requires Python 3.9, but would it be preferable to already to it now with `from __future__ import annotations`?

cbrnr (commented):
Looks better than expected 😄! I have a couple of questions (inline) plus one regarding the `py.typed` file – why is this empty file necessary (what does it do)? And if it is necessary, should it not be included as `package_data`?

cbrnr:
Also, what do I have to do to run the type checks? Is this somehow integrated into VS Code? Or do I need to run mypy manually? You also mentioned that Pyright still complains, I assume that's the checker that's activated when I switch "Python - Analysis - Type Checking Mode" to "basic", right? Is it possible to see an overview of all warnings (I only see them when opening a file)?

pyproject.toml
```diff
@@ -24,3 +24,14 @@ archs = 'x86_64 arm64'
...
 [tool.black]
 line-length = 92
+
+
+[tool.mypy]
+packages = ['sleepecg']
+exclude = '^sleepecg/test/'
+ignore_missing_imports = true
+disallow_untyped_defs = true
+no_implicit_optional = true
```
cbrnr, line 34: Is this still necessary? According to the [docs](https://mypy.readthedocs.io/en/stable/config_file.html#confval-implicit_optional), the argument is now called `implicit_optional`, which is `false` by default. So I think we don't need this at all since the default is what we want.

cbrnr:
Finally, I think we should add `.mypy_cache` to `.gitignore`.

sleepecg/io/physionet.py
hofaflo: According to the [Python docs](https://docs.python.org/3/library/typing.html#typing.Optional), `Optional` should be used if `None` is allowed. So it's appropriate in `_generate_gudb_md5`, but shouldn't be there in `download_physionet` and `_get_physionet_checksums`, I'll fix that!

sleepecg/io/gudb.py
hofaflo: Oh right, good point!

pyproject.toml
hofaflo: You're right!

hofaflo:
> Regarding the `py.typed` file – why is this empty file necessary (what does it do)? 

PEP561 [says so](https://peps.python.org/pep-0561/#packaging-type-information) :shrug:. I'm not totally sure but I assume it tells type checkers (such as mypy) that the package is typed.

> And if it is necessary, should it not be included as `package_data`?

Totally, it is included in the wheels and in the source archive (I guess because its parent directory is listed in MANIFEST.in?).

> I think we should add `.mypy_cache` to `.gitignore`.

For some reason my VS Code already ignored it so I didn't notice, but sure, I'll add it :D

cbrnr:
Looks great! Ready to merge?

hofaflo:
Yes!

cbrnr:
Thanks @hofaflo!

PR #106, opened 2022-08-01
Use Markdown links in docstrings

cbrnr:
This should look nicer in the rendered docs.


cbrnr:
It doesn't work: https://sleepecg--106.org.readthedocs.build/en/106/api.html

@hofaflo any idea if this can be done?

cbrnr:
I think this is the relevant issue: https://github.com/executablebooks/MyST-Parser/issues/228

cbrnr:
This will work with MkDocs in #119.

PR #107, opened 2022-08-02
Update benchmark figures

cbrnr:
The layout is a bit cleaner, and I've sorted the legend according to the actual detector performance at the longest signal duration. I'm also using the latest versions, which made `pyecg-pantompkins` significantly faster.
runtime__ltdb__2022_08_02__10_45_50

Current version: https://sleepecg.readthedocs.io/en/stable/heartbeat_detection.html#benchmarks


cbrnr:
Anything else we want to change? It is a bit unfortunate that `mne` and `neurokit2-nk` overlap almost completely, but I guess that's how it is...

I will do the other two figures next.

PR #112, opened 2022-08-03
Add MD5 checksums to GUDB

### LINKED ISSUE #111: `sleepecg.read_gudb()` always downloads data
Am I misunderstanding how our readers are supposed to work? I thought that datasets are only downloaded once if they are not available locally. However, it seems like this is not the case:

```python
import sleepecg

list(sleepecg.read_gudb())  # download if not already available in ~/.sleepecg/datasets
list(sleepecg.read_gudb())  # this should not trigger another download, but it does
```

The docstring states:

```
Required files are downloaded if not present in '<data_dir>/gudb'.
```

It also seems like this only affects `sleepecg.read_gudb()`.

cbrnr:
Fixes #111.


cbrnr:
@hofaflo let me know what you think. Since the checksums are not downloadable (I think), I just dropped them into a dict.

PR #116, opened 2022-08-09
Add function and method to plot ECG

### LINKED ISSUE #115: Add function (method) to plot ECG record
It would be nice if a given `ECGRecord` could be plotted, e.g. with a `.plot()` method. If the record contains an `annotation` field, these should also be plotted. In addition, it should be possible to plot custom markers (e.g. detected heartbeats), resulting in something like this:
Figure_1

(green = annotations, red = custom markers, i.e. detected heartbeats)

cbrnr:
Fixes #115.

An ECG trace can be plotted with the function `sleepecg.plot_ecg(ecg, fs)` or, if already available as an `ECGRecord`, with `record.plot()`. The record automatically plots the annotated heartbeats (which the function needs as an extra argument). Both variants support plotting an arbitrary number of annotations (e.g. to compare different detectors).


cbrnr:
I added `matplotlib` as a required dependency. Should we do that? Alternatively, we could also soft-import it inside the `plot_ecg()` function. But then how do I annotate the return type of this function without importing matplotlib? @hofaflo

cbrnr:
I solved this problem by returning `None` instead of `plt.Figure`. The only solution I've found is a stub file, but I really don't want to do this. We could add an `axis` parameter though, for people who want to plot on an existing axis.

cbrnr:
To Do:
- [ ] Add to docs

larsoner:
To fix rendering I would follow the MNE-Python config

https://github.com/mne-tools/mne-python/blob/2ba483ab91828be18a480e689d95f61c422ed6eb/doc/conf.py#L169-L178

With this folder structure of templates, which gets picked up by autosummary automatically (it looks for an `autosummary` folder inside your templates dir)

https://github.com/mne-tools/mne-python/tree/main/doc/_templates/autosummary

I don't think nested imports are so bad, but to avoid them usually creating a new file that the two problematic/cyclical files both need should work.

cbrnr:
Thanks @larsoner! I will take a look at how it's done in MNE-Python. However, I am tempted to try mkdocs first, which seems to be a lot easier in terms of configuration (and less effort for me than continue looking into Sphinx extension settings). We'll see how this works out.

Re cyclic imports, I think nested imports are OK for optional third-party dependencies, but they look weird for local modules. I've removed the plotting functionality from classification, so this should work now with standard imports. If that doesn't work (in the future), I'll create a new module with the problematic imports.

cbrnr:
Hmmm, not sure why it's still complaining about a cyclic import. I'm not importing anything from `plot` in `classification`.

cbrnr:
@hofaflo I managed to fix the circular import by moving some stuff to `utils.py`. Also, the docs now correctly cross-ref the `ECGRecord` methods. IMO this is ready for review. I know we discussed that our plotting functions should continue to return a `matplotlib.figure.Figure`, but maybe this can be added later (e.g. after you fixed the type annotations warnings)?

hofaflo:
Commits a575582, 376e8a4, 5145532, 6ea9560, and ee43db9 are not directly related to plotting ECG data. I'd prefer those changes to be made in a separate PR - could you remove them from this one?

> I know we discussed that our plotting functions should continue to return a `matplotlib.figure.Figure`, but maybe this can be added later (e.g. after you fixed the type annotations warnings)?

I think it would be best to add the return now, but add the annotation when fixing the typing issues.

cbrnr:
@hofaflo all of these commits were specific to Sphinx anyway. Now that we've switched to MkDocs, I think those changes are gone after rebasing. If you have time, please double-check – I will do that as well (and also continue to work on the open points).

docs/index.md
```diff
@@ -5,13 +5,8 @@ SleepECG provides tools for sleep stage classification when [EEG](https://en.wik
...
 
-## Overview
-- [Datasets](./datasets) shows all available datasets and instructions for retrieving NSRR data.
-- [Heartbeat Detection](./heartbeat_detection) demonstrates how to use the included heartbeat detector and shows benchmark results.
-- [Feature Extraction](./feature_extraction) lists all implemented HRV features.
-- [Classification](./classification) describes the included sleep stage classifiers and how to use the classification API.
-- [Configuration](./configuration) explains configuration settings in SleepECG.
-- [API](./api) has detailed information about all public functions and classes in SleepECG.
+## Documentation
+Documentation for SleepECG is available on [Read the Docs](https://sleepecg.readthedocs.io/en/stable/index.html).
```
hofaflo, line 9: I think this section isn't necessary on the docs' landing page.

docs/index.md
```diff
@@ -5,13 +5,8 @@ SleepECG provides tools for sleep stage classification when [EEG](https://en.wik
 - detecting heartbeats from ECG signals, and
 - classifying sleep stages (which includes preprocessing, feature extraction, and classification).
 
-## Overview
-- [Datasets](./datasets) shows all available datasets and instructions for retrieving NSRR data.
-- [Heartbeat Detection](./heartbeat_detection) demonstrates how to use the included heartbeat detector and shows benchmark results.
-- [Feature Extraction](./feature_extraction) lists all implemented HRV features.
-- [Classification](./classification) describes the included sleep stage classifiers and how to use the classification API.
-- [Configuration](./configuration) explains configuration settings in SleepECG.
-- [API](./api) has detailed information about all public functions and classes in SleepECG.
```
hofaflo, line 14: Why remove the overview?

docs/plotting.md
```diff
@@ -0,0 +1,38 @@
...
+
+```python
+from scipy.misc import electrocardiogram
+import sleepecg
+from sleepecg import ECGRecord
+
+ecg, fs = electrocardiogram(), 360
+beats = sleepecg.detect_heartbeats(ecg, fs)
+
+record = ECGRecord(ecg, fs, beats, id="scipy.misc.electrocardiogram()")
```
hofaflo, line 31:
```suggestion

ecg, fs = electrocardiogram(), 360
beats = sleepecg.detect_heartbeats(ecg, fs)

record = sleepecg.ECGRecord(ecg, fs, beats, id="scipy.misc.electrocardiogram()")
```

sleepecg/plot.py
```diff
@@ -0,0 +1,217 @@
...
+from .io.sleep_readers import SleepRecord, SleepStage
+from .utils import _merge_sleep_stages, _STAGE_INTS, _STAGE_NAMES, _time_to_sec
+
+
+def plot_ecg(
+    ecg: np.ndarray,
+    fs: float,
+    annotations: Optional[np.ndarray] = None,
+    *args: Optional[np.ndarray],
+    title: str = None,
```
hofaflo, line 21:
```suggestion
    title: Optional[str] = None,
```

hofaflo (commented):
The plots could get confusing when adding multiple additional annotations. Maybe they should be passed as a dictionary where the keys are used as legend entries?

cbrnr:
> The plots could get confusing when adding multiple additional annotations. Maybe they should be passed as a dictionary where the keys are used as legend entries?

I thought about that, but I wanted to be able to pass a NumPy array if there is just one sequence of annotations. I think this will be the case 99% of the time, so forcing users to pass a dict would be cumbersome.

Maybe we could clarify how to plot two sets of annotations with an example? I know the parameters are a bit unusual here, but in terms of calling the function it is very convenient.

docs/index.md
cbrnr: True. The reason I added it and at the same time removed the Overview (your next comment) is that the contents of `index.md` is identical to `README.md`. Which is easier to maintain, otherwise they might get out of sync quickly.

hofaflo: Ah I see! Then we could use `README.md` directly to avoid having to sync them manually?

hofaflo:
Good point... we could allow passing arbitrary kwargs (in addition to `*args`)? I think using the function isn't too complicated, I'm just sceptical of legend-less plots ;)

docs/index.md
cbrnr: Maybe... but how? Also, I'm not sure I want to include the badges in our docs, those should only be used in README.md. I think the simplest solution is to sync them manually (we can add a note in our CONTRIBUTING.md).

cbrnr:
Ah, you were talking about the legend, I didn't see that. Yes, if there are more than one sequences of annotations, there should be a legend with nice labels.

I guess we would have to move title before `**kwargs` then, right?

```
def plot_ecg(
    ecg: np.ndarray,
    fs: float,
    title: str = None,
    **kwargs: Optional[np.ndarray],
) -> None:
```

docs/index.md
hofaflo: I was hoping replacing `- Home: index.md` with `- Home: ../README.md` in `mkdocs.yml` would do the trick, but apparently that doesn't work, so nevermind 🤷.

hofaflo:
> I guess we would have to move title before `**kwargs` then, right?

I agree. The `Optional` isn't necessary for variable arguments though, it just means something can be `None` (see [here](https://docs.python.org/3/library/typing.html#typing.Optional)).

docs/index.md
cbrnr: OK, so let's sync them manually then.

cbrnr:
> I agree. The Optional isn't necessary for variable arguments though, it just means something can be None (see [here](https://docs.python.org/3/library/typing.html#typing.Optional)).

OK, so like this?

```
**kwargs: np.ndarray,
```

The provided type describes *one* kwarg, right?

hofaflo:
Exactly!

cbrnr:
What about the docstring description?

```
**kwargs : np.ndarray
```

or 

```
**kwargs : np.ndarray, optional
```

?

cbrnr:
To Do:

- [ ] Update example figures
- [ ] Add `plot_hypnogram()` to docs
- [ ] Anything else?

cbrnr:
@hofaflo I would like to stop returning a `Figure`, because this results in duplicated output in Jupyter notebooks (see the discussion [here](https://github.com/mne-tools/mne-python/issues/11165)).

I understand that the reason why you want to do is because users can then modify the generated figure afterwards if necessary. I see at least three options how to proceed:

1. Just return `None`.
2. Add a new parameter which accepts either a `Figure` or an `Axis`, which users can first generate e.g. by `fig, ax = plt.subplots()`, and plot on this object.
3. Add a new parameter `return_figure=False`, which switches between returning `None` or a `Figure`.

WDYT? Do you have other suggestions?

hofaflo:
I really want to allow users to modify (and save) created figures. While passing an existing (empty) `Figure` or `Axis` should be simple enough for `plot_ecg´, it would be less straightforward for `plot_hypnogram`. So to keep a consistent interface, I don't want to take this approach. This leaves two options which are both fine for me:

1. return a `Figure` and let users assign or use `;` to suppress the duplicated output
2. introduce the `return_figure` parameter you suggested above

cbrnr:
Then let's add the parameter.

cbrnr:
There's another option: the plot functions could return `fig, ax`. This prevents duplicate plots in notebooks. I tend to prefer this option, because we don't have to introduce a new parameter then. WDYT?

hofaflo:
> There's another option: the plot functions could return `fig, ax`.

That's perfect, very nice find!

cbrnr:
The only thing missing now is a short example for `plot_hypnogram()` in the docs. Will add that later.

cbrnr:
I think this looks good now! @hofaflo please feel free to merge if you are happy.

One thing I would like to change (in another PR) is that the content area (middle column) is pretty narrow. This is especially visible in the "Classification" section, where tables need to be scrollable. We could fix this by hiding the right sidebar (TOC), which I think is not needed.

docs/plot.md
```diff
@@ -0,0 +1,64 @@
...
+
+```python
+import sleepecg
+
+record = next(sleepecg.read_slpdb("*41*"))
+sleepecg.plot_hypnogram(
+    record,
+    record.sleep_stages,
+    "wake-rem-n1-n2-n3",
+    merge_annotations=True,
```
hofaflo, line 57: Is this parameter necessary even though `stages_mode` contains all five AASM stages?

cbrnr: You're right, it's not!

sleepecg/io/ecg_readers.py
```diff
@@ -42,10 +43,32 @@ class ECGRecord:
...
+
+    def plot(self, **kwargs: np.ndarray) -> None:
+        """
+        Plot ECG record.
+
+        Parameters
+        ----------
+        **kwargs : np.ndarray
+            Additional annotations to be plotted with different markers.
+        """
```
hofaflo, line 65: One last thing, I think `ECGRecord.plot`'s docstring should give the same information as `plot_ecg`'s:

```suggestion
    def plot(self, **kwargs: np.ndarray):
        """
        Plot ECG time series with optional markers.

        Parameters
        ----------
        **kwargs : np.ndarray
            Positions of annotations (i.e. heartbeats) in samples. If more than one marker
            sequence is given, the keywords will be used as labels in the plot legend.

        Returns
        -------
        fig : matplotlib.figure.Figure
            The figure.
        ax : matplotlib.axes.Axes
            The axes in the figure.
        """
```

cbrnr:
Done @hofaflo!

hofaflo:
Perfect, thanks @cbrnr!

PR #118, opened 2022-08-16
Trigger imports at heartbeat detector benchmark start

hofaflo:
Since #114. the detector packages are imported inside `detector_dispatch`, making the first call to the function take longer than subsequent ones. To keep runtimes similar, each detector should run once before the actual benchmark, triggering the required imports.


cbrnr:
Thanks @hofaflo!

PR #120, opened 2022-08-18
Remove zero-crossing check

cbrnr:
It could be that setting all values to zero before the first zero-crossing is not necessary (anymore). I'm currently running our benchmarks to see if it had any influence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - The detector benchmark example now requires only those packages that are actually used in the benchmark ([#114](https://github.com/cbrnr/sleepecg/pull/114) by [Clemens Brunner](https://github.com/cbrnr))
 - The Pan-Tompkins detector detects flat data at the beginning of a recording to avoid messing up its thresholds ([#87](https://github.com/cbrnr/sleepecg/pull/87) by [Raphael Vallat](https://github.com/raphaelvallat))
 - Switch documentation from Sphinx to MkDocs ([#119](https://github.com/cbrnr/sleepecg/pull/119) by [Clemens Brunner](https://github.com/cbrnr))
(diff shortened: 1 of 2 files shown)


hofaflo:
I'd like to find out what exactly is happening with those two records, but don't have access to my debugging-version of the detector for a few more weeks - so yes, let's leave it in for now.

cbrnr:
Sounds good! I will be unavailable for the next three weeks, so let's leave this PR open and continue discussing in September?

cbrnr:
@hofaflo do you have time to look into this?

hofaflo:
> @hofaflo do you have time to look into this?

Currently not :/

hofaflo:
I finally found time!

The records for which you found minor differences start during the rising edge of an R-peak, so this first R-peak is flattened and the corresponding T-wave is detected instead (for the following peaks everything works as it should).

`23_walking III` is a rather noisy signal whose peaks don't exhibit the typical morphology. A stray peak at the beginning leads to a low initialization of the average RR interval, which causes multiple searchbacks in the first 40s. Flattening the signal up to the first zero crossing removes this peak, leading to an initial `RR_average` which is closer to the truth, avoiding these searchbacks.

`11_maths chest` has a highly irregular morphology as well (very high T-waves and barely existing R-peaks. In both cases (set start to 0 or don't), the result is way off, detecting all the T-waves. Again, a small (incorrectly detected) peak at the start throws off the `RR_average` initialization, leading to lots of searchbacks.

To summarize:
- the major differences occur for records where the algorithm is not suited for the signal and produces nonsense anyway
- for the minor differences, removing the zero-crossing check is beneficial

So I'm in favor of merging this :)

cbrnr:
Awesome! I'll add a changelog and then merge this change! Thanks for checking!

PR #121, opened 2022-08-19
Fix GUDB checksum verification on windows

hofaflo:
In `gudb.GUDB_MD5`, filename keys use forward slashes. On windows, converting a `pathlib.Path` to a string produces backward slashes.


cbrnr:
Nice catch, thanks @hofaflo!

PR #123, opened 2022-11-05
Fix bug where setting user config values lets get_config retrieve the wrong key

hofaflo:
If a user config file with entries is present, the loop variable `key` overwrites the function argument. Checking all keys at once avoids the name clash.


cbrnr:
Good find! Can you maybe add a little test with such an invalid key?

hofaflo:
Sure, good idea!

hofaflo:
All config tests use a fixture which points the user config to a new temporary folder, so that shouldn't be a problem.

cbrnr:
Thanks @hofaflo!

PR #124, opened 2022-11-08
Add Python 3.11 support

cbrnr:
I *think* cibuildwheel should already build 3.11 wheels, let's see...

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,7 +31,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.8.1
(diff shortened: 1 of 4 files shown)


cbrnr:
We have to wait for Numba to support 3.11 (https://github.com/numba/numba/issues/8304).

hofaflo:
Tensorflow also doesn't support 3.11 yet, do you think we should add a note about that somewhere, e.g. put a disclaimer in the (yet to write) changelog entry?

Also, this worked locally, I'll try to find out why the CI still fails :thinking:

hofaflo:
Fixed, missed one of the `numba`s :sweat_smile:

cbrnr:
This is a very nice solution! Thanks @hofaflo!

PR #125, opened 2022-11-14
Hide TOC in docs

cbrnr:
This will (hopefully) create more space for the main content.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -37,6 +37,8 @@ theme:
       toggle:
         icon: material/brightness-4
         name: Switch to dark theme
+  hide:
+    - toc
 plugins:
   - mkdocstrings:
       handlers:


cbrnr:
Yes, this looks great. I'll tweak some column widths and then this should be good to go.

cbrnr:
@hofaflo this is already much better than before, so if you are happy please feel free to merge. I could not find out how to set column widths, so e.g. the first column in [this table](https://sleepecg--125.org.readthedocs.build/en/125/configuration/) is still too narrow. But in my opinion that's fine for now!

Next up I'll tackle the API sidebar TOC entries (there shouldn't be "Notes" or other unnecessary entries).

hofaflo:
Looks good! I would prefer to perform the aesthetic fixes in a separate PR, since this one's title suggests it's only about the TOC. Maybe you could move the two commits (3830a83 and 75cddf8) to #126 (and adapt its title to be even more general, e.g. "Fix various mkdocs issues")?

hofaflo:
Thanks @cbrnr!

PR #126, opened 2022-11-14
Fix various mkdocs issues

cbrnr:
- [ ] Add support for "Notes", "References", and "See Also" sections in griffe and Python handler repos (i.e. [here](https://github.com/mkdocstrings/griffe/blob/master/src/griffe/docstrings/numpy.py))
- [ ] Support cross-references user types in sections other than "Parameters" and "Returns" (e.g. it would be nice to have in the intro block).
- [x] Cross-reference external types (e.g. built-in or NumPy)


cbrnr:
I'll go ahead and merge, the other open issues need to be resolved upstream.

PR #134, opened 2022-12-10
remove dependency on pandas for read_gudb

hofaflo:
If I remember correctly, `read_gudb` uses `pandas` because its `load_csv` used to be a lot faster than numpy's `loadtxt`. This changed in numpy 1.23.0 when the function [was implemented in C](https://numpy.org/doc/stable/release/1.23.0-notes.html#faster-np-loadtxt), so I guess we can drop pandas as an optional requirement. I don't think it's necessary to bump the required numpy version to 1.23.0 since it is just a speed improvement, not a functional one, what do you think @cbrnr?


cbrnr:
I agree, no need to bump NumPy! Can you add a changelog entry?

cbrnr:
Thanks @hofaflo!

PR #139, opened 2023-01-17
Update SHHS encoding style and missing data handling

adgilbert:
I had some unicode decoding errors when reading SHHS data based on the windows "RIGHT SINGLE QUOTATION MARK" character. Changing the encoding to 'windows-1252' fixed this (as recommended in https://stackoverflow.com/a/40029793/3484157)

I also ran into some errors where gender or age values were missing for some rows in SHHS2 and I added the option for these parameters to be None.

I initially added logging for this, but removed it to avoid adding a new dependency (hence the multiple commits)


adgilbert:
@cbrnr please review when you can, thanks!

cbrnr:
Thanks for noticing and fixing these issues @adgilbert! 🚀

PR #144, opened 2023-04-13
Fix warnings during build time

### LINKED ISSUE #143: Build warnings
Building the conda-forge package in https://github.com/conda-forge/sleepecg-feedstock/pull/19#issuecomment-1506630898 revealed two warnings:

- [ ] There are two setuptools-related warnings for `sleepecg.classifiers` and `sleepecg.test`. These directories have been added automatically for now, but in the future this might not be the case. We could either add them manually in `setup.cfg` or `pyproject.toml` (https://setuptools.pypa.io/en/latest/userguide/package_discovery.html), or add empty `__init__.py` files in these directories.
- [ ] There are compiler warnings which we should probably fix:
    ```
      sleepecg/_heartbeat_detection.c: In function '_thresholding':
    sleepecg/_heartbeat_detection.c:475:45: warning: 'RR_high_limit' may be used uninitialized in this function [-Wmaybe-uninitialized]
      475 |                     if (RR_low_limit < RR_n && RR_n < RR_high_limit)
          |                         ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
    sleepecg/_heartbeat_detection.c:475:24: warning: 'RR_low_limit' may be used uninitialized in this function [-Wmaybe-uninitialized]
      475 |                     if (RR_low_limit < RR_n && RR_n < RR_high_limit)
          |                        ^
    sleepecg/_heartbeat_detection.c:330:33: warning: 'PEAKI' may be used uninitialized in this function [-Wmaybe-uninitialized]
      330 |                     SPKI = 0.25 * PEAKI + 0.75 * SPKI;
          |                            ~~~~~^~~~~~~
    ```

cbrnr:
This fixes `setuptools` warnings related to `sleepecg.classifiers` and `sleepecg.test`, which might not get added automatically to the package in the future.

In addition, I've reformatted `setup.cfg` to be as similar as possible to what it looks like when `setuptools` reformats it when building the source distribution.

Fixes #143.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -84,6 +84,8 @@ cibw =
 [options.packages.find]
 exclude =
     examples
+include =
+    sleepecg.classifiers
 
 
 [flake8]
--- a/setup.cfg
+++ b/setup.cfg
@@ -77,19 +77,3 @@ exclude =
     examples
 include =
     sleepecg.classifiers
-
-[flake8]
-extend-ignore = E203  # for Black compatibility
-max-line-length = 92
-max-doc-length = 92
-exclude =
-    ./build/*
-per-file-ignores = __init__.py:F401
-
-[pydocstyle]
-convention = numpy
-add_ignore = D105  # missing docstring in magic method
-
-[tool:pytest]
-markers =
-    c_extension: mark a test as belonging to a C extension
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,7 +21,6 @@ long_description = file: README.md
 long_description_content_type = text/markdown
 keywords = sleep, ecg, qrs, peak
 
-
 [options]
 include_package_data = True
 packages = find:
@@ -33,9 +32,7 @@ install_requires =
     scipy>=1.7.0
     tqdm>=4.60.0
 
-
 [options.extras_require]
-# complete package functionality
 full =
     joblib>=1.0.0
     matplotlib>=3.5.0
@@ -44,7 +41,6 @@ full =
     tensorflow>=2.7.0;python_version<'3.11'
     wfdb>=3.4.0
 
-# everything needed for development
 dev =
     black>=22.6.0
     flake8>=4.0.0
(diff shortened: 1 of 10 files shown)


cbrnr:
So we cannot make `sleepecg.test` a package, because this causes all tests to fail. Maybe this directory (as well as `sleepecg.classifiers`) shouldn't be a package after all? Please help, I have no idea what I'm doing 😆!

cbrnr:
@hofaflo do you know why making `classifiers` and `test` packages breaks the tests on cibuildwheel? Locally, all tests still pass with these changes...

cbrnr:
OK, got it, using `find_namespace:` instead of `find:` in `setup.cfg` solves this issue.

However, I still wonder about the error we got with `pytest` when these folders are true packages.

setup.cfg
```diff
@@ -33,9 +32,7 @@ install_requires =
     scipy>=1.7.0
     tqdm>=4.60.0
 
-
 [options.extras_require]
-# complete package functionality
```
hofaflo, line 38: could we keep the comments explaining what the individual extras_require sections are for?

cbrnr:
Sure, initializing with zero is fine!

Really weird stuff, why did you choose to make these two folders not proper packages in the first place? At least the test folder is very often a package with `__init__.py` from what I've seen in other projects.

setup.cfg
hofaflo: Ah I see! Yeah I'd prefer to keep them, would it be fine like this?
```ini
full = # complete package functionality
    joblib>=1.0.0
    matplotlib>=3.5.0
    mne>=1.0.0
    numba>=0.55.0;python_version<'3.11'
    tensorflow>=2.7.0;python_version<'3.11'
    wfdb>=3.4.0

```

hofaflo:
I dimly remember running into this exact issue with cibuildwheel/pytest and removing the `__init__.py` because of it :sweat_smile: 

For `classifiers`, I didn't think about making it a package because it doesn't contain any `.py`-files.

setup.cfg
cbrnr: Yes, that's fine! I'll add the comments.

cbrnr:
It's just the tests in `test_heartbeat_detection.py` that fail when we add an `__init__.py`:

```
  _____________________ test_squared_moving_integration_args _____________________
  
      def test_squared_moving_integration_args():
          """Test squared moving window integration argument parsing."""
  >       from sleepecg._heartbeat_detection import _squared_moving_integration
  E       ModuleNotFoundError: No module named 'sleepecg._heartbeat_detection'
  
  /project/sleepecg/test/test_heartbeat_detection.py:83: ModuleNotFoundError
```

All errors are related to `ModuleNotFoundError: No module named 'sleepecg._heartbeat_detection'` - why could that be happening? This file is a `.pyi` file, not sure if it has to do with that? It would be great if we could resolve this issue, because I have the feeling that it will hit us again in the future otherwise 😄.

PR #146, opened 2023-05-03
Silence TensorFlow messages

### LINKED ISSUE #145: Loading a classifier prints a lot of warnings
When loading a classifier that ships with SleepECG, a bunch of warnings are printed. I wonder if these warnings indicate real problems (in which case we should fix them) or if they can be ignored (in which case we should silence them).

WDYT @hofaflo?

```python
import sleepecg

clf = sleepecg.load_classifier("wrn-gru-mesa", "SleepECG")
```

I tried changing the logger level after we import `tensorflow`, but that didn't seem to do anything. Setting the environment variable [here](https://github.com/cbrnr/sleepecg/blob/f77e73b3bc790a535d36f44ee94290b0885c8bd8/sleepecg/classification.py#L295) (before the import) does get rid of all warnings except one message:

```python
import os
os.environ["TF_CPP_MIN_LOG_LEVEL"] = "2"
```

The remaining warning is an error:

```
2023-05-02 12:06:33.352056: E tensorflow/compiler/xla/stream_executor/cuda/cuda_driver.cc:266] failed call to cuInit: CUDA_ERROR_NO_DEVICE: no CUDA-capable device is detected
```

So setting the level to `"3"` gets rid of that one too.

cbrnr:
Fixes #145. I've added a new parameter in case anyone wants to keep the messages (but by default they are now silenced).

### DIFF
--- a/sleepecg/classification.py
+++ b/sleepecg/classification.py
@@ -296,7 +296,7 @@ def load_classifier(
 
         if classifier_info["model_type"] == "keras":
             import os
-            
+
             environ_orig = os.environ.copy()
             if silence_tf_messages:
                 os.environ["TF_CPP_MIN_LOG_LEVEL"] = "3"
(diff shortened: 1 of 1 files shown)


cbrnr:
@hofaflo please go ahead and merge if you are happy.

hofaflo:
Looks good, thanks @cbrnr!

PR #151, opened 2023-05-03
Fix warnings

cbrnr:
Trying to fix all warnings that occur in our tests. Not yet fully done, I need to ignore more DeprecationWarnings (that's the best we can do because they occur in external packages).


cbrnr:
I think it's pretty safe to merge, so I'll go ahead and do that. We can always revert in another PR if you disagree.

PR #152, opened 2023-05-06
Test detect_heartbeats with numba backend on python 3.11

hofaflo:
With numba now available for python3.11 (#147), we can stop skipping the detect_heartbeats test with numba backend.


cbrnr:
Thanks @hofaflo!

PR #179, opened 2023-05-24
Clean .gitignore

cbrnr:
I think we don't need to foresee every file we might want to ignore depending on which IDE or other tools someone uses. These should go into `.gitignore_global`, so currently only `*.so` remains (because this is part of the package, but should not be under source control).

For example, my `.gitignore_global` looks like this:

```
.DS_Store
.idea
.vscode
.venv
.envrc
.direnv
__pycache__
.mypy_cache
.pytest_cache
.ruff_cache
.ipynb_checkpoints
*.egg-info
```

@hofaflo WDYT?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,14 +1,1 @@
-**/__pycache__/
-!/.gitignore
-**/.ipynb_checkpoints/
-.idea
-.vscode
-.mypy_cache
-examples/datasets/
-examples/benchmark/results/
-build
-dist
-site
-*.egg-info
-*.pyd
 *.so
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,5 @@
 *.so
+.mypy_cache/
+.pytest_cache/
+.ruff_cache/
+__pycache__/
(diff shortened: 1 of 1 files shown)


cbrnr:
Alternatively, we could also argue that we should ignore all files that running/developing/... SleepECG produces, such as `.pytest_cache` (we're running Pytest when developing), `.mypy_cache` (same), `.ruff_cache` (same), etc.

I kind of prefer this option on second thought.

hofaflo:
I also prefer the second option. I'd argue that this should include a few more entries:
- `site`, which is created when building the docs locally
- `*.pyd`, which is the equivalent to `.so` on Windows
- `sleepecg.egg_info`, which appears when I run `pip install -e .`

cbrnr:
Done! I've excluded `sleepecg.egg-info` (with a `-` instead of an `_`), because that's what's being created on my macOS and Linux machines. If it's different on Windows please LMK.

hofaflo:
Ah yes, the `_` was a typo, perfect!

PR #182, opened 2023-06-28
Fix typos found by codespell

DimitriPapadopoulos:


cbrnr:
Nice, thanks @DimitriPapadopoulos! Maybe it would be worth adding Codespell to our pre-commit hooks.

DimitriPapadopoulos:
Why not, but be aware that codespell (or any spell checking tool for that matter) comes with false positives. Moreover, true or false positives may interfere with orthogonal attempts to commit.

cbrnr:
True, maybe a manual check once in a while is sufficient now that everything has been checked.

PR #192, opened 2023-09-25
Check if Python 3.12 works

cbrnr:
I don't expect it to work, but let's see.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,6 +111,7 @@ filterwarnings = [
     "error",
     'ignore:.*sample_rate:DeprecationWarning:pyedflib',
     'ignore:.*pkg_resources:DeprecationWarning',
+    'ignore:.*datetime.datetime.utcfromtimestamp\(\):DeprecationWarning',
 ]
 
 [tool.ruff]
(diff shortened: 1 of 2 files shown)


cbrnr:
It's not working due to Numba. Maybe we should not test against our optional dependencies? [Numba](https://github.com/numba/numba/issues/9197) and Tensorflow are usually slow to support the latest Python versions (for understandable reasons).

cbrnr:
@hofaflo what do you prefer?

1. Don't test on 3.12, but include Numba in our tests.
2. Test on 3.12, and remove Numba from our tests.

If we don't change anything, we're in option 1. I tend to prefer option 2 though, because Numba is only an optional dependency, and I'd like to support Python 3.12 (and any future releases) ASAP. In addition, we are also not testing Tensorflow functionality (also an optional dependency), which would also delay our support for latest Python releases, so we're already inconsistent in testing/not testing optional deps.

hofaflo:
I would like to keep the Numba test to make sure nothing breaks there, but skip it for now, like we did last year (see #124)

cbrnr:
It looks like we will have to wait until Python 3.12 is released, unless we want to enable pre-release dependencies (e.g. `cffi`). I say we just wait.

cbrnr:
Seems like we need to wait for https://github.com/dateutil/dateutil/issues/1284 (they have already fixed it, but have not cut a new release yet). Or we just silence this warning for now and hopefully not forget about it.

Issue #16, opened 2021-08-15
Migrating to Python 3.9


hofaflo:
When we require [Python 3.9](https://docs.python.org/3/whatsnew/3.9.html) in the future, we can make use of the following updates:

- [ ] [Type hinting generics](https://www.python.org/dev/peps/pep-0585/) (in every function expecting or returning e.g. a list, dict or tuple).

If you encounter situations for which Python 3.9 would allow a more elegant solution, please note them here.

hofaflo:
Since we started using `from __future__ import annotations` in #105, there isn't really much left to do here. I think we can close this issue - and if a situation arises where supporting 3.8 actually makes something difficult we can discuss again, WDYT @cbrnr?

cbrnr:
We could leave it open though to remind us to remove the future import once we move to 3.9. Or would we get some kind of info message from some linter?

hofaflo:
I was hoping mypy would say something, but apparently it doesn't :/ So yes, let's leave it open.

cbrnr:
When do you think we should drop support for Python 3.8? I would argue that supporting the last three versions is more than sufficient. This would mean 3.9, 3.10, and 3.11 – but of course only if all features work on all versions. Therefore, I suggest to drop support for 3.8 once Numba and Tensorflow are compatible with 3.11.

hofaflo:
I don't have a strong opinion on this, though I don't see why we shouldn't support 3.8 as long as it doesn't hinder development :thinking:

cbrnr:
True, but it's good to have a set of rules in place. If we take https://devguide.python.org/versions/ as a guide, we could keep 3.8 around until October 2024. But I don't think we have to. Fewer supported versions means fewer maintenance burden. I don't know. I thought it would be nice to have a rule instead of deciding ad-hoc "this version is already too old so let's drop support for it".

hofaflo:
Okay, I'm fine with "support the latest 3 python versions for which all features work"

cbrnr:
We only need to wait for Numba and Tensorflow to support Python 3.11, then we can drop Python 3.8:

https://github.com/numba/numba/issues/8304
https://github.com/tensorflow/tensorflow/issues/58032

EwoutH:
[...]

PR #195, opened 2023-11-10
Use edfio for reading and writing edf files

hofaflo:
[edfio](https://github.com/the-siesta-group/edfio) could replace PyEDFlib and mne for our use cases. We would have to drop support for Python 3.8, but that is in line with the schedule we decided on in #16 anyway.


docs/examples.md
```diff
@@ -86,27 +86,27 @@ plt.show()
...
-raw.crop((start - raw.info["meas_date"]).seconds, (stop - raw.info["meas_date"]).seconds)
-
-# get ECG time series as 1D NumPy array
-ecg = raw.get_data().squeeze()
+start = datetime(2023, 3, 1, 23, 0, 0)
+stop = datetime(2023, 3, 2, 6, 0, 0)
+rec_start = datetime.combine(edf.startdate, edf.starttime)
+edf.slice_between_seconds((start - rec_start).seconds, (stop - rec_start).seconds)
+
+# get ECG time series and sampling freqeuncy
```
cbrnr, line 107:
```suggestion
# get ECG time series and sampling frequency
```

cbrnr (commented):
This is great! Thank you for this PR, I've only find a typo!

cbrnr:
Really? We need `pooch` just to get the SciPy ECG toy data?

hofaflo:
Apparently yes :grimacing: Do you prefer to bundle an ecg signal with the package instead?

cbrnr:
> Do you prefer to bundle an ecg signal with the package instead?

Yes. I'd store it as compressed `.npz`, then it takes up only 163kB. We could store it in `sleepecg/data` (in analogy to `sleepecg/classifiers`)?

cbrnr:
Maybe we could even add a little function to access the toy data (`sleepecg.get_toy_ecg()` or something similar)?

cbrnr:
Regarding dropping support for Python 3.8, I don't think we can already remove `from __future__ import annotations`, right? This is still required for Python 3.9, or is it?

cbrnr:
Of course the other option is to export the toy data to EDF and add `edfio` to our base requirements.

hofaflo:
I opted for the .npz file because the .edf would be a bit larger and introduce slight signal deviations.

Ad future imports: A few have become obsolete, but most of them have to stay until 3.10, yes.

cbrnr:
Great! I assume nothing has to be done to make sure the data gets added to the wheels, correct? We're still using `MANIFEST.in` unfortunately.

hofaflo:
I think `graft sleepecg` makes sure this is included, yes. As cibuildwheel runs pytest after installing the built wheel, the checks would fail otherwise anyway.

cbrnr:
Thanks @hofaflo!

PR #197, opened 2023-11-10
Fix sdist

### LINKED ISSUE #196: Tweak MANIFEST.in
I think we should decide what the source distribution should contains. There are basically two options:

1. All files required to install the package
2. All files under source control

Currently, we're mixing these things, because we're including e.g. `.github/` and `readthedocs.yml`, but not `mkdocs.yml` or `docs/`. WDYT?

cbrnr:
Fixes #196. I decided to include all files under source control, because (1) it is easier, (2) it's probably a good thing to have an archive (snapshot) of all sources for every release, and (3) I think that's what we originally wanted to do (the missing `docs` folder was actually a bug, because we included a `doc` folder, which does not exist because we renamed it to `docs` at some point but forgot to adapt `MANIFEST.in`).

So hopefully, you are OK with this sudden change of mind. If so, please feel free to merge.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,5 @@
 graft .github
-graft doc
+graft docs
 graft examples
 graft sleepecg
 global-exclude *.so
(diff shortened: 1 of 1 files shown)


hofaflo:
Fine with me, thanks @cbrnr!

Issue #14, opened 2021-08-12
Tests should not use `_heartbeat_detection`


cbrnr:
The tests still import the C implementation directly. We should use the public function instead and perform the tests with all three backends.

This raises the additional question: how do I build the C extension locally?

hofaflo:
I'm with you on performing tests on all backends. I'd keep the tests for type- and valuechecks of the C extension though, to make sure possible future updates don't mess something up there. We could [mark them](https://docs.pytest.org/en/6.2.x/example/markers.html#marking-test-functions-and-selecting-them-for-a-run), disable them by default and only run them if required?

For building the C extension locally I usually use `pip install -e .[full]` (`[full]` makes sure `wfdb` and `numba` are installed, as those are required for testing).  `python setup.py build bdist_wheel` followed by `pip install <path-to-wheel>` should work as well.

cbrnr:
> I'm with you on performing tests on all backends. I'd keep the tests for type- and valuechecks of the C extension though, to make sure possible future updates don't mess something up there. We could mark them, disable them by default and only run them if required?

Of course, I didn't mean to imply that we should drop the C extension tests. However, we could use the public function with `backend='c'` to perform these tests (and then add tests for the other backends if necessary).

> For building the C extension locally I usually use `pip install -e .[full]` (`[full]` makes sure `wfdb` and `numba` are installed, as those are required for testing). `python setup.py build bdist_wheel` followed by `pip install <path-to-wheel>` should work as well.

Thanks, I'll go with `pip install -e .[full]` then.

hofaflo:
#18 marks the module containing the C extension tests, so they can be excluded by calling `pytest -m "not c_extension"`.

PR #199, opened 2023-11-29
Heart rate data

MashRiza:
I have introduced Changes to the SleepECG package, primarily focusing on enabling sleep stage classification using heart rate data when EEG signals are not available. Key features of this update include:

Heart Rate to R-Interval Conversion:

Implemented an algorithm to convert heart rate data into R-intervals. This crucial development allows the use of heart rate data, which is more readily available in many scenarios, for sleep stage classification. Handling Data Gaps:

Added robustness to the system by integrating code that effectively handles scenarios where data might be missing for extended periods. This ensures more reliable performance and resilience in real-world applications. Web-Based and Local Application:

Originally developed for a web-based project, these enhancements are equally applicable for local deployments, thereby broadening the usability scope of the SleepECG package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,34 @@ from sleepecg import detect_heartbeats, get_toy_ecg
 ecg, fs = get_toy_ecg()  # 5 min of ECG data at 360 Hz
 beats = detect_heartbeats(ecg, fs)  # indices of detected heartbeats
 ```
-
+### Example if Heart Rate data is available
+data = [
+    {'TimeStamp': '1690409900', 'HeartRate': '55'},
+    {'TimeStamp': '1690410000', 'HeartRate': '60'},
+    {'TimeStamp': '1690410120', 'HeartRate': '65'},
+    {'TimeStamp': '1690410300', 'HeartRate': '62'},
+    {'TimeStamp': '1690410600', 'HeartRate': '59'},
+    {'TimeStamp': '1690410900', 'HeartRate': '58'},
+    {'TimeStamp': '1690411200', 'HeartRate': '61'},
+    {'TimeStamp': '1690411320', 'HeartRate': '56'},
+    {'TimeStamp': '1690411380', 'HeartRate': '52'},
+    {'TimeStamp': '1690411440', 'HeartRate': '50'},
+    {'TimeStamp': '1690411500', 'HeartRate': '56'},
+    {'TimeStamp': '1690411560', 'HeartRate': '58'},
+    {'TimeStamp': '1690411620', 'HeartRate': '60'},
+    {'TimeStamp': '1690411680', 'HeartRate': '65'},
+    {'TimeStamp': '1690412740', 'HeartRate': '60'},
+    {'TimeStamp': '1690412800', 'HeartRate': '62'},
+    {'TimeStamp': '1690413800', 'HeartRate': '62'},
+    {'TimeStamp': '1690514000', 'HeartRate': '62'}
+]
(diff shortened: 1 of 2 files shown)


MashRiza:
Thanks for taking the time to review my pull request. In this update, I focused on enhancing the SleepECG package's capability to classify sleep stages using heart rate data (a heart rate with corresponding timestamp )when EEG signals are unavailable.
example data :
data = [
    {'TimeStamp': '1690409900', 'HeartRate': '55'},
    {'TimeStamp': '1690410000', 'HeartRate': '60'},
    {'TimeStamp': '1690410120', 'HeartRate': '65'} ...]
Specifically, I introduced an algorithm to convert heart rate data into R-intervals, providing a valuable addition to the existing features. While I acknowledge that RRI computation is already part of SleepECG, my contribution lies in showcasing a practical example in the example folder. This example illustrates how to calculate R-intervals from heart rate and perform subsequent sleep stage classification, catering to scenarios where only heart rate data is accessible.

Additionally, I addressed the robustness of the system by implementing code to handle missing data for extended periods. This improvement aims to ensure more reliable performance in real-world applications.

Considering the ongoing discussion in #14 about robustness, I'll certainly try my best to chime in to contribute to the discourse and provide further insights into the implemented changes.

Moreover, these enhancements were initially developed for a web-based project, but I made sure they are equally applicable for local deployments, thereby expanding the SleepECG package's usability scope .

cbrnr:
It looks like the main addition is an option to convert RRI back to time stamps, correct? I think this could be quite useful, but in order to integrate it into SleepECG, you should put it into a dedicated function. I'm not sure if it belongs to the `heartbeats` or `feature_extraction` module, probably in the former. The function should receive an array of RRIs and return an array of heart beats.

Please remove your changes from `README.md`. Also, the example needs to be adapted in order to use the dedicated function. We can decide later how to best create an example (currently I think it is too long).

@hofaflo WDYT?

hofaflo:
If I understand correctly, the input data would not be RR intervals, but an irregularly sampled heart rate. But I generally agree, a function which takes heart rate and corresponding timestamps as input and produces (estimated) heartbeat times which can then be used as input to the existing feature extraction functions would be a useful addition!

cbrnr:
Definitely! RRI is the inverse of the HR, so the function would support both measures.

MashRiza:
> Definitely! RRI is the inverse of the HR, so the function would support both measures.

The idea for this usage came up as I was working on a current project. I analyze heart rate data gathered from smart bands, and I need to deploy it on the web using JSON files as input. Thanks for your review; I will make the changes you mentioned earlier (wrap it in dedicated function and delete changes in README.md) and ask for more help if it's possible. I also recommend checking a competition on Kaggle, which has a relatively useful topic.
kaggle.com/competitions/child-mind-institute-detect-sleep-states

MashRiza:
> > Definitely! RRI is the inverse of the HR, so the function would support both measures.
> 
> The idea for this usage came up as I was working on a current project. I analyze heart rate data gathered from smart bands, and I need to deploy it on the web using JSON files as input. Thanks for your review; I will make the changes you mentioned earlier (wrap it in dedicated function and delete changes in README.md) and ask for more help if it's possible. I also recommend checking a competition on Kaggle, which has a relatively useful topic. kaggle.com/competitions/child-mind-institute-detect-sleep-states

i have added HeartRateSleep function as last function in feature_extraction.py . 
the function also has some codes for handling missing intervals as it is common in smart band/watch .
i would love to read your reviews

PR #219, opened 2024-05-20
Switch to src layout

cbrnr:
Maybe - probably - it will just work?


cbrnr (commented):
Yeah, OK, it seems like it doesn't "just work"...

cbrnr (commented):
It's working, but since we include "examples" and "tests" in the wheels, these are considered packages and can be imported. We don't really want that, so we need to either find a fix or move these directories inside the `sleepecg` package...

cbrnr:
BTW, this has always been the case, so it's not something introduced with this PR...

cbrnr:
I think the best way forward is to not bundle the tests with the package. Regarding the examples, I'm not sure what to do. The scripts to generate the built-in classifiers are pretty important, so I wonder if they should go into the package (but if so, where?). The benchmark could also go into a separate subpackage (e.g. `sleepecg.benchmark`), but then the question is how to configure it (since it currently uses a `config.yml`). @hofaflo do you have any opinion on this?

cbrnr (commented):
I went with the simplest option and removed both the tests and examples. They are available in the source distribution anyway.

PR #224, opened 2024-06-17
Pin NumPy < 2.0 for now

cbrnr:
We need to support NumPy >= 2.0 in our build system, or otherwise pin to NumPy < 2.0. Let's see what we can do.


cbrnr:
Until Tensorflow supports NumPy 2.0 (https://github.com/tensorflow/tensorflow/issues/67291), I think we should pin to < 2.0 in the meantime...

PR #229, opened 2024-07-19
Fix some type checker warnings

cbrnr:
WIP


cbrnr:
@hofaflo would you care to take a look, specifically if you are OK with the config change?

cbrnr:
I think `get_config_key()` should be renamed to `get_config_value()` (thanks @hoechenberger!). I'll do that.

hofaflo:
Yes, I think this is a very good idea, thanks for fixing that and the typing chaos! :sweat_smile:

PR #241, opened 2024-10-09
Add support for Python 3.13

cbrnr:
In addition, bump dependencies.

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -90,10 +90,8 @@ jobs:
       - run: ls -al dist
       - name: Triage dependencies
         run: |
-          if [[ "${{ matrix.python-version }}" != "3.13" ]]; then
-            echo "PIP_EXTRA=wfdb numpy<2 numba pytest-error-for-skips" | tee -a $GITHUB_ENV
-            echo "PYTEST_EXTRA=--error-for-skips" | tee -a $GITHUB_ENV
-          fi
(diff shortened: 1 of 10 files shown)


cbrnr (commented):
NumPy errors are all in wfdb, I'll try to fix them (will link the corresponding PR once it's up).

cbrnr (commented):
We have to wait for https://github.com/MIT-LCP/wfdb-python/pull/511 (and a new release).

cbrnr:
This seems to work nicely. However, I don't like that some dependencies are installed in `cibuildwheel.yml` since everything should already be defined in `pyproject.toml`. Specifically, we created the `cibw` extras group exactly for this purpose. @larsoner do you know how the "Triage dependencies" step can be modified to use this extras group instead of manually installing packages?

larsoner:
In theory you could modify the `python -m pip install ./dist/*.whl` command, probably first by getting the wheel filename then using the `file://` syntax to get to something like `pip install file://$WHEEL_PATH[some_new_option]` and add `some_new_option` to `pyproject.toml` or whatever. TBD whether or not `pip` works properly with a `file://` plus optional specifier, though!

cbrnr:
Ah, you're right, the `cibuildwheel` tests are still there, but they only test the wheel they've just built (i.e., using Python 3.10 currently).

It's still a bit of duplication going on because of the additional tests, which is not ideal. I'll try and see if dropping `cibuildwheel` (and instead building and testing wheels on runners using `uv`) simplifies things in a future PR.

PR #249, opened 2024-12-13
Activity count feature addition

simon-p-2000:
Added the option to download, process and read the activity counts from the mesa dataset. This is done by extending the read_mesa function. The activity counts represent processed accelerometer data and can be used as an additional feature to train the existing classifiers.

The option to download and process the activity counts is integrated into the read_mesa function by a variable activity_source and is set to False by default. As such, unless specified, no activity counts will be used. If either cached activity counts or the actigraphy csv files in the mesa database are used, the SleepRecord which the function returns, contains an additional array with the activity counts for the specified recording. 

This new feature is integrated in such a way that it works for both the offline and non offline option provided by the read_mesa function. Since the actigraphy csv files are much smaller than the EDF files, no option to delete the csv files afer downloading them is provided.


src/sleepecg/io/sleep_readers.py
```diff
@@ -220,6 +236,13 @@ def read_mesa(
             f"possible options: {heartbeats_source_options}"
         )
 
+    activitiy_source_options = {None, "cached", "actigraphy"}
```
cbrnr, line 239:
```suggestion
    activity_source_options = {None, "cached", "actigraphy"}
```

src/sleepecg/feature_extraction.py
```diff
@@ -56,8 +56,13 @@
         "LF_HF_ratio",
     ),
     "metadata": ("recording_start_time", "age", "gender", "weight"),
+    "actigraphy": ("activity_counts", "dummy_feature"),
+}
+_FEATURE_ID_TO_GROUP = {
+    id: group
+    for group, ids in _FEATURE_GROUPS.items()
+    for id in (ids if isinstance(ids, tuple) else (ids,))
```
cbrnr, line 64: I don't understand why this is necessary?

tests/test_sleep_readers.py
```diff
@@ -143,6 +194,55 @@ def test_read_mesa(tmp_path):
...
+
+    assert len(records) == 2
+
+    for rec in records:
+        assert rec.sleep_stage_duration == 30
+        assert set(rec.sleep_stages) - valid_stages == set()
+        assert len(rec.activity_counts) == 4
+        assert os.path.exists(
+            f"{tmp_path}/mesa/preprocessed/activity_counts/{rec.id}-activity-counts.npy"
+        )
```
cbrnr, line 220: Can you use `pathlib` instead?

examples/classifiers/wrn_gru_mesa.py
```diff
@@ -27,7 +27,9 @@
 if TRAIN:
     print("‣  Starting training...")
     print("‣‣ Extracting features...")
-    records = list(read_mesa(offline=False))
+    records = list(
+        read_mesa(offline=False, records_pattern="000*", activity_source="actigraphy")
```
cbrnr, line 31: I guess this should revert to all patterns once you have confirmed that it works with all files, right?

cbrnr:
I have made two cosmetic commits (please check if you're OK with the changes). Can you rebase please before applying my suggestions?

examples/classifiers/wrn_gru_mesa.py
```diff
@@ -15,9 +15,9 @@
     set_nsrr_token,
 )
 
-set_nsrr_token("your-token-here")
+set_nsrr_token("YOUR TOKEN HERE")
 
-TRAIN = True  # set to False to skip training and load classifier from disk
+TRAIN = False  # set to False to skip training and load classifier from disk
```
cbrnr, line 20: These changes should probably also reverted?

simon-p-2000: reset it in the following commit

examples/classifiers/wrn_gru_mesa.py
simon-p-2000: I reverted it in the following commit

tests/test_sleep_readers.py
simon-p-2000: I implemented this via pathlib in the following commit

src/sleepecg/feature_extraction.py
simon-p-2000: I found a different way to do it and forgot to remove it until now

tests/test_sleep_readers.py
```diff
@@ -12,10 +12,42 @@
...
 from edfio import Edf, EdfSignal
 
-from sleepecg import SleepStage, get_toy_ecg, read_mesa, read_shhs, read_slpdb
+from sleepecg import (
+    SleepStage,
+    get_toy_ecg,
+    read_mesa,
+    read_shhs,
+    read_slpdb,
+)
```
cbrnr, line 21: These imports could be in a single line, or no?

simon-p-2000: Fixed in the next commit

cbrnr:
Yep, this looks good! Can you run the other two as well and compare with https://sleepecg.readthedocs.io/en/stable/classification/?

simon-p-2000:
wrn_gru_mesa_weighted:
grafik

simon-p-2000:
grafik
wrn_gru_mesa classifier

It appears that the classifiers are working similarly to before the addition of the activity count feature, so when the activity counts are not added as a feature the classifier is not affected. The next step would be to run the classifiers again with the activity counts added as a feature and see if that improves the classifier performance.

docs/datasets.md
```diff
@@ -59,6 +59,14 @@ set_nsrr_token("<your-download-token-here>")
 mesa = read_mesa(records_pattern="00*")  # note that this is a generator
 ```
 
+For the mesa dataset SleepECG also supports downloading and  storing the so-called activity counts. These are metrics
+that quantify movement of the subject via the recording and processing of 3D accelerometer measurements. They are
+calculated via a proprietary algorithm in the Actiware software by Phillips. In order to include activity counts in the 
+datasets downloaded from the mesa database, the [`read_mesa()`](sleepecg.read_mesa) function needs to be called with the
+parameter *activity_source* set either to *'actigraphy'* to download the data outright or *'cached'* to use cached
+activity counts if they were already downloaded previously.
```
cbrnr, line 67:
```suggestion
SleepECG supports downloading and storing activity counts for the MESA dataset. These metrics quantify a subject's movement based on accelerometer measurements recorded and processed using a proprietary algorithm in Philips Actiware. To access activity counts, call [`read_mesa()`](sleepecg.read_mesa) with `activity_source='actigraphy'` to download the data or `activity_source='cached'` to use previously downloaded counts.
```

tests/test_sleep_readers.py
```diff
@@ -144,6 +194,55 @@ def test_read_mesa(tmp_path):
...
+            activity_source="cached",
+        )
+    )
+
+    assert len(records) == 2
+
+    for rec in records:
+        assert rec.sleep_stage_duration == 30
+        assert set(rec.sleep_stages) - valid_stages == set()
+        assert len(rec.activity_counts) == 6
```
cbrnr, line 243: Why is the expected number of `rec.activity_counts` different in the cached vs. non-cached tests?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [UNRELEASED] - YYYY-MM-DD
 ### Added
 - Add support to store NSRR token in environment variable or user config ([#243](https://github.com/cbrnr/sleepecg/pull/243) by [Simon Pusterhofer](https://github.com/simon-p-2000))
+- Add download and storage of activity_counts for mesa data ([#249](https://github.com/cbrnr/sleepecg/pull/249) by [Simon Pusterhofer](https://github.com/simon-p-2000))
```
```suggestion
- Add support for downloading and storing activity counts for the MESA dataset ([#249](https://github.com/cbrnr/sleepecg/pull/249) by [Simon Pusterhofer](https://github.com/simon-p-2000))
```

tests/test_sleep_readers.py
cbrnr: Hm, where exactly is that difference? You are calling `_create_dummy_mesa()` with the exact same arguments?

simon-p-2000: I added docstrings, are they understandable?

src/sleepecg/io/sleep_readers.py
```diff
@@ -350,13 +420,86 @@ def read_mesa(
...
+                end_line = (
+                    int(
+                        next(
+                            row["line"]
+                            for row in activity_data
+                            if row.get("linetime") == recording_end_time_str
+                        )
+                    )
+                    - 1
+                )
```
cbrnr, line 486: I feel like finding the first occurrence of a matching line using `next()` is not immediately obvious. Also, what happens if no match is found? Can this case be ruled out (i.e., does it always find a match)?

Maybe a good old loop would be easier to read, such as:

```python
for row in activity_data:
    if row.get("linetime") == recording_end_time_str:
        end_line = int(row["line"]) - 1
        break
else:
    print(f"Skipping {record_id} due to missing line matching {recording_end_time_str}.")
    continue
```

Depending on whether or not there can be a "no match" case, the `else` branch could also be removed.

simon-p-2000: I processed about 1700 actigraphy files and never had the problem that there was no match, although it might be possible. I will add the else statment to cover the edgecase and rewrite it with a loop

cbrnr: In this case, do you not want to use the `else` belonging to the `for` loop? I think this is a nice use case here, as the `else` is only hit if the `for` loop exhausts all iterations, i.e. the `break` statement was never hit.

simon-p-2000: I wasn't aware of this for... else structure, it sounds very useful here. I will implement it

cbrnr:
Thanks @simon-p-2000!

PR #251, opened 2025-01-14
ENH: abi3

### LINKED ISSUE #230: Consider providing abi3 wheels
I think it would be very useful if SleepECG provided [abi3 (stable limited ABI) wheels](https://cibuildwheel.pypa.io/en/stable/faq/#abi3), because these are forward-compatible, which means that in theory we only have to provide one wheel (e.g. `cp39-abi3`) and it will work with all future versions.

### LINKED ISSUE #235: Python 3.13 roadmap
We cannot enable Python 3.13 currently, because

- [ ] Numba does not support it yet (https://github.com/numba/numba/issues/9413) (there is no release yet)
- [ ] We have to remove the NumPy < 2 pin, because there won't be 3.13 wheels for the old NumPy 1.26 (and building from source is not an option)
- [ ] At least one of our dependencies pins NumPy to < 2 (https://github.com/MIT-LCP/wfdb-python/pull/494)

larsoner:
Closes #230
Closes #235

1. Use abi3 mode
2. Bump macos-14 to -15
3. Build native wheels on macOS in CIs
4. Test built wheels on a separate Python (esp. useful now that they'll be abi3)
5. Add `aarch64` builds (hopefully fast enough?)
6. Use `abi3audit` to check the wheels
7. Deduplicate test and release into cibuildwheel
8. Add codecov

Worth bumping to 3.10+ at the same time?


larsoner:
At least locally:
```
$ pip wheel --no-deps .
$ pip install abi3audit
$ abi3audit --strict -v sleepecg-0.6.0.dev0-cp39-abi3-linux_x86_64.whl 
[10:41:26] 💁 sleepecg-0.6.0.dev0-cp39-abi3-linux_x86_64.whl: 1 extensions scanned; 0 ABI version mismatches and 0 ABI violations found
```

.github/workflows/release.yml
```diff
@@ -1,63 +0,0 @@
...
-    name: Upload to PyPI
-    needs: [style, build_wheels, build_sdist]
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/download-artifact@v4
-        with:
-          pattern: artifact-*
-          merge-multiple: true
-          path: dist
-      - uses: pypa/gh-action-pypi-publish@v1.12.3
```
larsoner, line 63: This was ~80% duplicated with `tests.yml` so I folded both into a `cibuildwheel.yml`. Don't love the name but maybe good enough. Could use templates or something to deduplicate but it's not as readable or easy to work with.

larsoner:
Okay I think this one is good to go @cbrnr

cbrnr:
This is great 🚀, thank you so much @larsoner! I only have a couple of questions just so I understand everything you did here:

- Apparently, the abi3 wheels now work with Python 3.9+, but will this now correctly show up in the badge at the top of `README.md`?
- How does codecov work? I don't see it in the checks, does it have to be enabled? Is codecov and coveragerc related?
- I assume once wfdb supports NumPy >= 2, some of the conditions can be removed/simplified, right?
- Is the `numpy < 3` bound necessary?
- I've migrated most of my projects to `uv` – would you say that this setup would also benefit from switching (if so, it can be done later, not in this PR)?

larsoner:
> Apparently, the abi3 wheels now work with Python 3.9+, but will this now correctly show up in the badge at the top of README.md?

I just changed openmeeg and it has:



A bit odd actually that it doesn't say `3.10+` or similar since it *does* have a lower bound but that must be a PyPI badge limitation (which isn't surprising given the limited abi3 adoption so far).

> How does codecov work? I don't see it in the checks, does it have to be enabled? Is codecov and coveragerc related?

Yes all related, you might need to configure something on https://app.codecov.io/gh/cbrnr/sleepecg but I'd wait until something landed in main. And you'll want to take the token from there and add it as a GitHub secret as CODECOV_TOKEN.

> I assume once wfdb supports NumPy >= 2, some of the conditions can be removed/simplified, right?

Yeah I think you should be able to install all dependencies in all runs rather than triaging based on NumPy/Python version.

> Is the numpy < 3 bound necessary?

Maybe not, but given the issues with NumPy 2.0 rollout it seems safest. I thought they recommended it at one point but I can't find a record of it, so it could be removed.

> I've migrated most of my projects to uv – would you say that this setup would also benefit from switching (if so, it can be done later, not in this PR)?

No idea, I've actually never used uv. But @mscheltienne has and I think liked it


> Regarding bumping to 3.10+, I'd say if SleepECG now runs on 3.13, I'm fine with dropping support for 3.9, but I'll leave it up to you.

Okay, not totally necessary. There were some additions to the stable API for 3.10 that I thought you might need as well but it doesn't look like it!

mscheltienne:
I use `uv` daily, but mostly for env management. I saw they added build, publication, etc.. but I don't consider this part stable enough yet, at least last I checked.

cbrnr:
I'm already using the build and publish parts of uv (together with all the rest), it works perfectly! I'll make a follow up PR to switch to uv, this should simplify some things.

cbrnr:
Thanks @larsoner! I just realized that you did not add a changelog entry. Do you want to add something? I think this is a pretty important change, since Python 3.13 is now supported.

larsoner:
Sure, do you want to just add it to #253 ? If not I can make a quick PR

cbrnr:
Please make a quick PR, otherwise I might forget again!

cbrnr:
BTW, maybe worth trying? https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

PR #252, opened 2025-01-14
MAINT: Trivial CI bump

larsoner:
Workflows require approval until a PR has been merged (unless you tweak this setting), so get a tiny PR in first before putting in the real work and iteration in https://github.com/cbrnr/sleepecg/pull/251, otherwise the approval process will be a limiting factor


cbrnr:
Thanks @larsoner!

PR #253, opened 2025-01-16
Add coverage config

cbrnr:
Move `.coveragerc` to `pyproject.toml` and try to automatically run it with `pytest`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,11 +105,7 @@ repair-wheel-command = [
 
 [tool.coverage.run]
 branch = true
-source = ["sleepecg"]
-omit = [
-    "*/examples/*",
-    "*/setup.py",
-]
+source = ["src/sleepecg"]
 
 [tool.mypy]
 ignore_missing_imports = true


cbrnr:
Hmmm, `pytest` runs fine locally, I don't know why it's failing here (apparently, `pytest-cov` does not get installed?).

larsoner:
You'll need to add `pytest-cov` to https://github.com/cbrnr/sleepecg/blob/94e83901502aafb545f6e0b4c39d435ccf54b837/pyproject.toml#L78

pyproject.toml
```diff
@@ -116,6 +121,7 @@ filterwarnings = [
     'ignore:.*datetime.datetime.utcfromtimestamp\(\):DeprecationWarning',
     'ignore:(?s).*Pyarrow:DeprecationWarning',
 ]
+addopts = "--cov"
```
larsoner, line 124: From looking at lost of tracebacks on CIs and locally I'd recommend a couple more options we have in MNE-Python
```suggestion
addopts = "--cov -rfEXs --cov-report= --tb=short"
```
some of which could then be removed from https://github.com/cbrnr/sleepecg/blob/94e83901502aafb545f6e0b4c39d435ccf54b837/.github/workflows/cibuildwheel.yml#L98

PR #256, opened 2025-01-22
Run tests with uv

cbrnr:
Maybe this can replace cibuildwheel, we'll see.


cbrnr:
No idea what's going on with the numpy import error on Windows though...

cbrnr:
See #274 and #275.

PR #262, opened 2025-03-03
Adding new classifiers, extending feature extraction

simon-p-2000:
Added new classifiers that use sklearn
Extended feature extraction to accomodate actigraphy data
Added new function to transform features into shape suitable for sklearn


simon-p-2000:
Additional things to be implemented:
add ability to save/load sklearn classifiers
add NN classifiers that uses torch (discuss, perhaps this is better in a separate PR)

src/sleepecg/classification.py
```diff
@@ -13,6 +13,7 @@
 from typing import Any, Protocol
 from zipfile import ZipFile
 
+import joblib
```
cbrnr, line 16: If you want to use `joblib` to save/load sklearn models, we'd need to add it to the required dependencies. But I'm not sure that we should do that. You could just use `pickle`, which is essentially doing the exact same thing minus some efficiency benefits (which we probably do not need anyway).

Thinking further, maybe ONNX would be the best option, because it is format-independent, which we could also use for our Keras models. I think we'll need ONNX anyway for converting our Keras models to PyTorch models.

However, as you say, we should probably tackle the Keras -> PyTorch conversion in a separate PR. For now, I suggest that you use `pickle` instead of `joblib` here.

src/sleepecg/feature_extraction.py
```diff
@@ -657,6 +657,10 @@ def _extract_features_single(
             )
         elif feature_group == "metadata":
             X.append(_metadata_features(record, num_stages))
+        elif feature_group == "actigraphy":
+            if record.activity_counts is not None:
+                activty_counts = record.activity_counts
```
cbrnr, line 662: Typo, this should be `activity_counts`.

src/sleepecg/feature_extraction.py
```diff
@@ -368,10 +368,10 @@ def _hrv_frequencydomain_features(
...
     hf_mask = (0.15 < freq) & (freq <= 0.4)
 
-    total_power = trapezoid(psd[:, total_power_mask], freq[total_power_mask])
-    vlf = trapezoid(psd[:, vlf_mask], freq[vlf_mask])
-    lf = trapezoid(psd[:, lf_mask], freq[lf_mask])
-    hf = trapezoid(psd[:, hf_mask], freq[hf_mask])
+    total_power = np.trapz(psd[:, total_power_mask], freq[total_power_mask])
+    vlf = np.trapz(psd[:, vlf_mask], freq[vlf_mask])
+    lf = np.trapz(psd[:, lf_mask], freq[lf_mask])
+    hf = np.trapz(psd[:, hf_mask], freq[hf_mask])
```
cbrnr, line 374: `np.trapz` is deprecated, please use `np.trapezoid` instead.

src/sleepecg/feature_extraction.py
```diff
@@ -57,6 +56,7 @@
         "LF_HF_ratio",
     ),
     "metadata": ("recording_start_time", "age", "gender", "weight"),
+    "actigraphy": ("activity_counts", "dummy_feature"),
```
cbrnr, line 59: Did you forget to remove `dummy_feature`, or is this needed?

src/sleepecg/classification.py
```diff
@@ -22,6 +23,54 @@
...
+        A 1D array containing the annotated sleep stage for each sample. The sleep stages
+        are merged based on the stages_mode parameter.
+    record_ids : np.ndarray
+        A 1D array containing a calculated index for each valid sample that is returned.
+    """
+    record_ids = np.hstack([i * np.ones(len(X)) for i, X in enumerate(features)])
+    features_stacked = np.vstack(features)
+    stages_stacked = np.hstack(_merge_sleep_stages(stages, stages_mode))
+    features_stacked[np.isinf(features_stacked)] = -1
+    features_stacked[np.abs(features_stacked) > 1e9] = -1
```
cbrnr, line 68: Why this clamp above 1e9?

simon-p-2000: I will have to look for a proper value, for now I chose an arbitrary large value.

cbrnr: But why is this necessary at all? What's the problem with leaving such large values in the data?

src/sleepecg/feature_extraction.py
cbrnr: Do you mean you need a tuple? Then you can use a tuple with a single item like this (note the trailing comma):

```python
"actigraphy": ("activity_counts",),
```

src/sleepecg/feature_extraction.py
simon-p-2000: I have no idea why this has changed, I will rewert it to its original state.

src/sleepecg/classification.py
simon-p-2000: It might not actually be necessary since some kind of normalization is used with either classifier type

src/sleepecg/feature_extraction.py
simon-p-2000: This leads to an error, I will try debugging it

src/sleepecg/feature_extraction.py
simon-p-2000: fixed

src/sleepecg/classification.py
simon-p-2000: using pickle now

src/sleepecg/classification.py
simon-p-2000: removed the masking for values above 1e9

src/sleepecg/io/sleep_readers.py
```diff
@@ -420,7 +420,7 @@ def read_mesa(
 
         parsed_xml = _parse_nsrr_xml(xml_filepath)
 
-        activity_counts = None
+        activity_counts: None | list[int] | np.array = None
```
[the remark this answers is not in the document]
simon-p-2000: Removed the type annotations

src/sleepecg/feature_extraction.py
simon-p-2000: fixed

cbrnr:
Thanks @simon-p-2000!

PR #263, opened 2025-03-12
Add sklearn classifiers

simon-p-2000:
Adding a variety of sklearn classifiers along with pre trained classifiers.

Added function for preparing and reshaping features for use in sklearn classifiers. Adding functionality to save and load sklearn classifiers.


simon-p-2000:
Added example pytorch classifier. However, the performance is somewhat poor

src/sleepecg/classification.py
```diff
@@ -146,7 +146,7 @@ def prepare_data_pytorch(
     mask_value: int = -1,
 ) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
```
cbrnr, line 147: I think the type hints need to be adjusted.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [UNRELEASED] - YYYY-MM-DD
 ## Added
 - Add support for activity counts feature ([#262](https://github.com/cbrnr/sleepecg/pull/262) by [Simon Pusterhofer](https://github.com/simon-p-2000))
+- Adding sklearn classifiers along with necessary infrastructure to save/load them ([#262](https://github.com/cbrnr/sleepecg/pull/263) by [Simon Pusterhofer](https://github.com/simon-p-2000))
```
```suggestion
- Added support for Scikit-Learn and PyTorch models ([#263](https://github.com/cbrnr/sleepecg/pull/263) by [Simon Pusterhofer](https://github.com/simon-p-2000))
```

docs/classification.md
```diff
@@ -18,7 +28,7 @@ A weaker weighting approach is likely required to find the optimal middle ground
 
 ![wrn-gru-mesa confusion matrix](./img/wrn-gru-mesa.svg)&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;![wrn-gru-mesa-weighted confusion matrix](./img/wrn-gru-mesa-weighted.svg)
 
-
+Additional classifiers using the Sklearn library have been added. The classifiers denoted with the 'actigraphy' suffix used the activity counts feature during training.
```
cbrnr, line 31:
```suggestion
Additional classifiers using [Scikit-Learn](https://scikit-learn.org/stable/) can be trained with the example scripts available in the `examples/classifiers` folder.

```

src/sleepecg/classification.py
```diff
@@ -15,6 +15,7 @@
 from zipfile import ZipFile
 
 import numpy as np
+import torch
```
cbrnr, line 18: `torch` is an optional dependency so please do not import it globally. If you need it within a function, you should also import it in the function. If you need it for type annotations, guard the import with

```python
if TYPE_CHECKING:
```

src/sleepecg/classification.py
simon-p-2000: adjusted the type hints

src/sleepecg/classification.py
simon-p-2000: How should I deal with it in the classifier files? The dependency is used throughout the entire file there

cbrnr: In the classifier examples I think it's OK to globally import it, as these files are not part of the SleepECG package.

simon-p-2000: Understood, I edited the classification.py file according to the example in plot.py

PR #283, opened 2025-08-02
Remove wheel from build-system.requires

DimitriPapadopoulos:
Frontend build tools that support PEP 517 install `wheel` automatically if needed.

Also require a recent setuptools for PEP 639 support.


cbrnr:
Thanks @DimitriPapadopoulos!

PR #285, opened 2025-08-02
Keep configuration lean: remove `license-files`

DimitriPapadopoulos:
Most build backends will pick up all `LICEN[CS]E*` files.

That's of course true of setuptools:
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-files

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,6 @@ name = "sleepecg"
 version = "0.6.0.dev0"
 description = "A package for sleep stage classification using ECG data"
 license = "BSD-3-Clause"
-license-files = ["LICENSE"]
 authors = [
     {name = "Florian Hofer", email = "hofaflo@gmail.com"},
     {name = "Clemens Brunner", email = "clemens.brunner@gmail.com"},


DimitriPapadopoulos:
See https://github.com/pypa/packaging.python.org/issues/1854.

DimitriPapadopoulos:
PEP 639 is recent enough for some backends to still have issues with `license-files`. That's actually the case for [maturin](https://github.com/PyO3/maturin) (https://github.com/astral-sh/ruff/pull/19599#issuecomment-3139658776).

I haven't tested `uv-build`. [The uv build backend](https://docs.astral.sh/uv/concepts/build-backend/#the-uv-build-backend) documentation states that:
> The uv build backend is a great choice for most Python projects. It has reasonable defaults, with the goal of requiring zero configuration for most users [...]

I would therefore expect it to have appropriate defaults for `license-files`.

How about we wait until the switch to `uv-build` and reconsider this PR then?

cbrnr:
I'm pretty sure the `license-files` field is required by `uv-build` in order to include the file in the sdist: https://docs.astral.sh/uv/concepts/build-backend/#file-inclusion-and-exclusion

I think it's also more explicit, and having the two fields doesn't hurt either. If we remove it now I can almost guarantee that I'll forget to add it when switching to `uv-build`, so I'd rather keep it around.

cbrnr:
> I've tested `uv build` and indeed it doesn't pick up `LICENSE` by default, without `license-files = ["LICENSE"]`

Do you think this is worth reporting? Or rather, asking if they want to add some defaults? I'm not sure, I kind of prefer to be explicit, so without an entry in `license-files`, it makes a lot of sense that it doesn't include a file.

DimitriPapadopoulos:
https://github.com/astral-sh/uv/issues/15040

Issue #267, opened 2025-05-07
Use agcounts package for actigraphy counts feature


cbrnr:
I just found the agcount package (https://github.com/actigraph/agcounts), which we could (should?) use instead of our own implementation. Or at the very least, I think we should validate our implementation against this package.

viranovskaya:
Thanks for pointing this out. I checked the current implementation and the merged MESA work in #249.

As far as I can tell, SleepECG is not converting raw accelerometer samples to activity counts here; it reads the already processed Philips Actiware `activity` values from the MESA files. `agcounts.get_counts()` expects raw X/Y/Z acceleration, so it cannot directly replace or validate the current reader using the same MESA CSVs.

If the intended scope is to support raw accelerometer input, I could prepare a small validation fixture using the reference data provided by `agcounts`, while keeping the existing MESA reader separate. Is that the direction you had in mind?

cbrnr:
Thanks for the heads up, I wasn't aware of this difference. I think we should then just compare our implementation to theirs (starting from the activity values). Depending on how complex it is to process accelerometer data and compute activity values, I'm not sure if we should use this package or implement it ourselves.

viranovskaya:
Thanks. I checked the current path again: SleepECG does not calculate activity counts here. `read_mesa()` reads the proprietary Actiware `activity` column, and `extract_features()` passes those values through unchanged. `agcounts` starts from raw X/Y/Z acceleration, so there is no like-for-like comparison starting from the MESA activity values.

I see two separate scopes:

1. keep the MESA path as a reader and alignment test for existing Actiware values;
2. if raw accelerometer support is wanted, add a separate path backed by `agcounts` and validate it on their reference data.

I would avoid implementing a second count algorithm ourselves. Would it make sense to handle these as two separate cases?

cbrnr:
Yes, I think this makes sense! Raw accelerometer support would be nice, as currently I don't think there's a way to use that data for sleep staging other than the MESA datasets, right? I haven't worked on this repository for a long time, so I can't really remember and I don't have time to dive into it at the moment...

viranovskaya:
I checked the current main branch. There is already a generic path for non-MESA data: users can calculate activity counts externally, pass a one-dimensional array as `SleepRecord(activity_counts=...)`, and include `actigraphy` in `feature_selection`.

[...]

PR #315, opened 2026-07-30
docs(actigraphy): document external activity counts

viranovskaya:
## Why

`SleepRecord` already accepts externally calculated activity counts, but the workflow for non-MESA data was not documented. Invalid or misaligned counts could also reach feature extraction without a clear error.

## What Changed

- document an external `raw X/Y/Z -> agcounts -> SleepRecord` workflow
- explain that activity-count methods are device-specific and should remain consistent between training and prediction
- note that the bundled classifiers do not currently use actigraphy
- validate that `activity_counts` is one-dimensional and aligned with the feature-extraction epochs
- add tests for valid, missing, multidimensional, and misaligned counts

## Verification

The targeted tests confirm that valid external counts pass through unchanged, while missing, multidimensional, and length-mismatched inputs raise clear errors.

Refs #267.

### DIFF
--- a/docs/feature_extraction.md
+++ b/docs/feature_extraction.md
@@ -78,6 +78,46 @@ Group identifier: `metadata`
 
 Group identifier: `actigraphy`
 
-| Feature           | Description                                                                                            |
-|-------------------|--------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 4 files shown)


cbrnr:
Thanks @viranovskaya!

PR #317, opened 2026-08-04
Update dev dependencies

DimitriPapadopoulos:
Since mypy and ruff are run by pre-commit, I would recommend also removing these lines:
https://github.com/cbrnr/sleepecg/blob/75b915a9842a7ec9173722ac479685a1ff5963a5/pyproject.toml#L61
https://github.com/cbrnr/sleepecg/blob/75b915a9842a7ec9173722ac479685a1ff5963a5/pyproject.toml#L65

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,13 +1,13 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.12.2
+    rev: v0.16.1
     hooks:
       - id: ruff-check
         args: [ --fix ]
       - id: ruff-format
 
   - repo: https://github.com/pre-commit/mirrors-mypy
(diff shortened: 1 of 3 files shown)


cbrnr:
> Since mypy and ruff are run by pre-commit, I would recommend also removing these lines:

Yes, feel free to remove them. I'd also replace pip with uv in the instructions, similar to https://github.com/cbrnr/mnelab/blob/main/CONTRIBUTING.md - WDYT?

DimitriPapadopoulos:
Done. CI tests appear to be failing because https://physionet.org is down.

DimitriPapadopoulos:
As for switching the instructions to `uv`, I would rather address that in a different PR, as there are multiple files to modify and check.

cbrnr:
Thanks @DimitriPapadopoulos!

Issue #141, opened 2023-02-24
ECG peak detection gets stuck when data is very noisy


raphaelvallat:
Hi,

This is an issue I have encountered a few times already — when the ECG data is very noisy for either part of or most of the recording, the `detect_heartbeats` function just gets stuck forever. This is problematic when you run peak detection across hundreds of participants and cannot visually check each recording.

**To reproduce the error:**

Download the EDF file with 8 hours ECG: https://drive.google.com/file/d/1ReYlFPAd3-dYk0C2WF7nDRQS3MxE2lyE/view?usp=share_link

```python
import mne
from sleepecg import detect_heartbeats

# Load ECG
raw = mne.io.read_raw_edf("original_ecg.EDF", preload=True, verbose=False)
ecg = raw.get_data(units="uV")[0]
sf = raw.info["sfreq"]# Load ECG

# Peak detection
detect_heartbeats(ecg, sf) # gets stuck
```

In that case, the ECG data is so bad that honestly it would be better not to attempt the peak detection at all...!

**Potential solutions:**

1. Modification of the algorithm to avoid getting stuck when no peak is found?
2. Pre-screening of the ECG signal quality — if the data is very noisy then the detection is skipped and an empty array is returned instead? For example we could use some kind of ECG signal quality metric (see `method="zhao2018"` in https://neuropsychology.github.io/NeuroKit/functions/ecg.html#ecg-quality) 

 Thanks,
Raphael

cbrnr:
Hi @raphaelvallat! This is probably the biggest issue with our R peak detector, and I actually suggest that we implement *both* solutions you mention.

Re 1, I assume that the algorithm keeps lowering the thresholds in cases where there are no peaks in the data. Although I thought that we limited this to a fixed number, maybe I'm wrong or there is a problem in our implementation. Can you specify where the detector gets stuck when it tries to find peaks in such bad data?

I am also not sure how easy it is to just skip a detection, we'd have to give it a try. Naively, I'd stop finding peaks after a certain number of tries (i.e. a certain number of times the thresholds were lowered) and go to the next segment. But since segments are defined by peaks, this might require some changes.

Re 2, does this method assess signal quality continuously? Or do you get one evaluation score for the entire signal? We could think about adding a `mask` parameter, which tells the algorithm which segments are valid and which segments should be ignored. Again, this will require some adaptations to make the detector work with such kind of data.

PR #319, opened 2026-08-04
Fix repeated searchback scans on unusable ECG segments

viranovskaya:
Addresses the detector-stall part of #141.

The detector could appear to hang on a long noisy interval because every unsuccessful searchback started again after the previous detected peak. The same increasingly long interval was scanned up to 15 times at every new sample.

This change keeps track of the part of the interval that has already been checked. After an unsuccessful searchback, the next pass only checks newly added samples. The thresholds do not change while this branch is active, so rescanning the older part cannot produce a new candidate.

The cursor resets whenever a heartbeat is detected. The same logic is used by the Python, Numba and C implementations.

Checks:

- added a deterministic regression test for repeated searchback scans;
- added a noisy-signal test for all three backends;
- full test suite: 55 passed;
- Ruff and `git diff --check` passed;
- existing toy and seeded signals return unchanged heartbeat indices.

This fixes the computational stall. It does not add ECG quality classification or a mask for unusable intervals.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Remove support for macOS on Intel architecture (x86_64) ([#301](https://github.com/cbrnr/sleepecg/pull/301) by [Clemens Brunner](https://github.com/cbrnr))
 
 ### Fixed
-- Avoid repeated searchback scans when heartbeat detection encounters long unusable ECG intervals
+- Avoid repeated searchback scans when heartbeat detection encounters long unusable ECG intervals ([#319](https://github.com/cbrnr/sleepecg/pull/319) by [Daria Agafonova](https://github.com/viranovskaya))
 
 ## [0.5.9] - 2025-02-01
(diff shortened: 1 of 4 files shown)


cbrnr:
Very nice, can you confirm that the file linked in #141 works with this PR?

viranovskaya:
Yes. I downloaded the exact `original_ecg.EDF` file linked in #141 and ran the original reproduction code against this PR.

The recording contains 31,732,000 samples at 1000 Hz, or about 8.81 hours. The default C backend completed in 1.09 seconds and returned 32,056 peaks. I also checked the Numba and Python backends. They completed in 2.12 and 40.74 seconds, respectively, and all three returned identical peak indices.

cbrnr:
Very nice! Can you fix the style error and if this is ready convert the draft PR to a full PR?

cbrnr:
Thank you so much @viranovskaya 🚀!

PR #321, opened 2026-08-05
Add CAP Sleep Database reader

### LINKED ISSUE #35: Add functions for reading annotated sleep data
The goal is to have functions which download and read all common polysomnography datasets. Each record will be stored in a dataclass containing heartbeat times and annotated sleep stages.

Each reader should be added via a separate PR, this issue gives an overview of the progress.

Datasets:
- [x] [MESA](https://sleepdata.org/datasets/mesa) (Multi-Ethnic Study of Atherosclerosis) #28 
- [x] [SHHS](https://sleepdata.org/datasets/shhs) (Sleep Heart Health Study) #48 
- [x] [SLPDB](https://physionet.org/content/slpdb/1.0.0/) (MIT-BIH Polysomnographic Database) #47
- [ ] [MASS](http://ceams-carsm.ca/en/mass/) (requires registration including project description and ethics approval, so this is not really open)
- [ ] [CAPSLPDB](https://physionet.org/content/capslpdb/1.0.0/) (CAP Sleep Database)
- [ ] [CinC 2018](https://physionet.org/content/challenge-2018/1.0.0/) (You Snooze You Win - The PhysioNet Computing in Cardiology Challenge 2018)
- [ ] [sleep-accel](https://physionet.org/content/sleep-accel/1.0.0/) (Motion and heart rate from a wrist-worn wearable and labeled sleep from polysomnography)
- [ ] [DREAMS](https://zenodo.org/record/2650142)
- [ ] [ISRUC-SLEEP Dataset](https://sleeptight.isr.uc.pt/)

Planned future extensions:
- metadata
- actigraphy data

viranovskaya:
Closes #35.

This adds `read_capslpdb()` for the CAP Sleep Database. The reader:

- uses the REMlogic `Event` field for sleep stages and maps S3 and S4 to N3;
- aligns 30-second annotations with the EDF recording and handles missing internal rows;
- supports EKG, ECG, ECG1-ECG2, and separate ECG1/ECG2 channels;
- warns and skips `n16` because it does not contain an ECG channel;
- supports cached heartbeats and the existing PhysioNet download options.

I checked the parser against all 108 public annotation files, containing 108,720 valid sleep-stage rows. `nfle27` has one malformed row, which is skipped with a warning. I also ran the complete reader on `n12`, producing 990 sleep-stage epochs and 30,656 heartbeat times.

The tests cover stage mapping, time and header variants, missing rows, EDF boundaries, midnight alignment, malformed rows, irregular grids, bipolar ECG, heartbeat rebasing, `n16`, and downloaded EDF cleanup.

### DIFF
--- a/src/sleepecg/io/capslpdb.py
+++ b/src/sleepecg/io/capslpdb.py
@@ -1,0 +1,334 @@
+# © SleepECG developers
+#
+# License: BSD (3-clause)
+
+"""Read records from the CAP Sleep Database."""
+
+from __future__ import annotations
+
+import csv
+import datetime
+import warnings
+from collections.abc import Iterator
+from itertools import pairwise
+from pathlib import Path
+from typing import TYPE_CHECKING, NamedTuple
+
+import numpy as np
+
+if TYPE_CHECKING:
+    from edfio import Edf
+
+from sleepecg.config import get_config_value
+from sleepecg.heartbeats import detect_heartbeats
+from sleepecg.io.physionet import _list_physionet, download_physionet
+from sleepecg.io.sleep_readers import SleepRecord, SleepStage
+
+
+class _ParseCapslpdbResult(NamedTuple):
+    sleep_stages: np.ndarray
+    scoring_start_offset: int
+    malformed_rows: int
+
+
+_CAPSLPDB_STAGE_MAPPING = {
+    "SLEEP-S0": SleepStage.WAKE,
+    "SLEEP-S1": SleepStage.N1,
+    "SLEEP-S2": SleepStage.N2,
+    "SLEEP-S3": SleepStage.N3,
+    "SLEEP-S4": SleepStage.N3,
+    "SLEEP-REM": SleepStage.REM,
+    "SLEEP-MT": SleepStage.UNDEFINED,
+}
+
+
+def _clock_seconds(value: str) -> int:
+    parsed = datetime.datetime.strptime(value.replace(".", ":"), "%H:%M:%S")
+    return parsed.hour * 3600 + parsed.minute * 60 + parsed.second
+
+
+def _time_seconds(value: datetime.time) -> int:
+    return value.hour * 3600 + value.minute * 60 + value.second
+
+
+def _normalise_capslpdb_header(value: str) -> str:
+    return value.strip().replace("Duration [s]", "Duration[s]")
+
+
+def _read_capslpdb_stage_rows(
+    annotation_filepath: Path,
+) -> tuple[list[tuple[int, SleepStage]], int]:
+    lines = annotation_filepath.read_text(encoding="ascii").splitlines()
+    try:
+        header_index = next(
+            index
+            for index, line in enumerate(lines)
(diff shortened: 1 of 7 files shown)


cbrnr:
Could you rebase to pull in some new Ruff rules please?

viranovskaya:
Done, I rebased the branch onto the current main and updated the code for the new Ruff rules.

Local checks pass: Ruff, formatting, mypy, and the full test suite with 68 tests.

cbrnr:
Very nice! Just one thing that should be checked: if `heartbeats_source="cached"`, the reader should not download the EDF file(s), but instead use the cached heartbeats (like the other readers). Here's what Claude found:

`heartbeats_source="cached"` still re-downloads all EDFs. In `read_capslpdb`, `download_physionet(..., extensions=[".edf", ".txt"], ...)` runs unconditionally for every requested record before the per-record loop, regardless of `heartbeats_source`. So even in `"cached"` mode, `read_edf(edf_filepath, ...)` needs a local EDF, meaning a later run with `"cached"` will silently re-download the whole ~40 GB dataset just to read starttime/duration — defeating the point of caching heartbeats + `keep_edfs=False`.

Compare `read_mesa`/`read_shhs`: there the EDF is only downloaded inside the `"ecg"` branch, so `"cached"` never touches it.

Suggested fix: cache `recording_start_time`/`duration` alongside the heartbeats `.npy` when running with `"ecg"`, so `"cached"` mode never needs the EDF at all.

viranovskaya:
Thanks for catching this. I updated the reader so cached mode downloads only the annotation files and never downloads or reads an EDF. The ECG path now stores the recording start time and duration alongside the cached heartbeat array, and cached mode restores them from that metadata.

I also added a regression test that fails if cached mode attempts to download or read an EDF. Ruff, formatting, mypy, and the full test suite pass locally: 69 tests.

src/sleepecg/io/capslpdb.py
```diff
@@ -0,0 +1,375 @@
...
+                metadata_filepath,
+                recording_start_time,
+                recording_duration,
+            )
+            if not edf_was_available[record_id] and not keep_edfs:
+                edf_filepath.unlink()
+
+        heartbeat_times = heartbeat_times[
+            (heartbeat_times >= 0) & (heartbeat_times < scored_duration)
+        ]
```
cbrnr: I think the second masking could be conditional, right? Otherwise, it happens twice for the `"ecg"` case (which isn't wrong but just unnecessary).

```suggestion
        else:
            heartbeat_times = heartbeat_times[
                (heartbeat_times >= 0) & (heartbeat_times < scored_duration)
            ]
```

viranovskaya: Right, thanks. I made the second mask conditional, so it now runs only for cached heartbeats. Ruff, formatting, mypy, the 10 CAPSLPDB tests, and the full 69-test suite all pass.

cbrnr:
Perfect! Thanks @viranovskaya!

PR #324, opened 2026-08-05
Change to default line width in ruff

### LINKED ISSUE #320: Use 88 characters line width
SleepECG should use the default of 88 characters per line. I don't know why I started with 92, but now I think we should just follow what everyone else is doing.

DimitriPapadopoulos:
Use the default width of 88 characters instead of 92.

Fixes #320.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,7 +109,7 @@ markers = ["c_extension"]
 filterwarnings = ["error"]
 
 [tool.ruff.lint]
-select = ["C4", "D", "E", "F", "FURB", "I", "PERF", "W", "UP"]
+extend-select = ["C4", "D", "FURB", "I", "PERF", "W", "UP"]
 ignore = ["D105"]
 exclude = ["setup.py"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,12 +110,12 @@ filterwarnings = ["error"]
 
 [tool.ruff.lint]
 extend-select = ["C4", "D", "FURB", "I", "PERF", "W", "UP"]
-ignore = ["D105"]
-exclude = ["setup.py"]
+ignore = ["D105", "DTZ"]
 
 [tool.ruff.lint.per-file-ignores]
-"__init__.py" = ["F401"]
+"__init__.py" = ["F401", "F403"]
 "**/examples/*.py" = ["D100"]
+"setup.py" = ["D10"]
 
 [tool.ruff.lint.pydocstyle]
 convention = "numpy"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,13 +108,10 @@ pretty = true
 markers = ["c_extension"]
 filterwarnings = ["error"]
 
-[tool.ruff]
-line-length = 92
-exclude = ["setup.py"]
-
 [tool.ruff.lint]
 select = ["C4", "D", "E", "F", "FURB", "I", "PERF", "W", "UP"]
 ignore = ["D105"]
+exclude = ["setup.py"]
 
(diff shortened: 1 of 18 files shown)


cbrnr:
Thanks @DimitriPapadopoulos!

pyproject.toml
```diff
@@ -109,7 +109,7 @@ markers = ["c_extension"]
 filterwarnings = ["error"]
 
 [tool.ruff.lint]
-extend-select = ["C4", "D", "FURB", "I", "PERF", "W", "UP"]
+extend-select = ["C4", "D", "PERF", "W"]
```
DimitriPapadopoulos, line 112: @cbrnr Not 100 % sure `FURB`, `I` or `UP` should be removed. From https://github.com/scientific-python/cookie/pull/843:
> Ruff 0.16 turns on 413 rules without configuration. Compared the default set against the full rule index: `BLE`, `DTZ`, `FA`, `FLY`, `INT`, `PIE`, and `YTT` are fully covered (including preview rules), and _I_ is covered except for `I002`, which needs the `lint.isort.required-imports` setting. No other group is fully covered.
> [...]
> * Removed the `RF102` check ("isort must be selected"). `RF101` (`B`) and `RF103` (`UP`) stay, because those groups are not fully default.

But then the defaults may be good enough for us.

cbrnr: OK, I thought these were included by default? https://docs.astral.sh/ruff/default-rules/

DimitriPapadopoulos: Not all rules, but I guess the ruff authors made a sensible choice.

Nevertheless, I'd like to understand which rules in `B` and `UP` they left out, and why they left out [`I002`](https://docs.astral.sh/ruff/rules/missing-required-import/) — it should be a no-op anyway without [`required-imports`](https://docs.astral.sh/ruff/settings/#lint_isort_required-imports).

cbrnr: OK, I think the defaults should be fine, but feel free to add specific rules that you think are important.

henryiii: Here's the B and UP rules that are not selected by default still:

:robot: _AI text below_ :robot:

Checked against ruff 0.16.2 (`--show-settings` enabled list vs `ruff rule --all`).

**`B` (flake8-bugbear): 43 rules, 29 on by default, 14 not.**

Stable rules you only get by selecting `B`:

| Code | Rule                                        |
| ---- | ------------------------------------------- |
| B007 | unused-loop-control-variable                |
| B011 | assert-false                                |
| B024 | abstract-base-class-without-abstract-method |
| B027 | empty-method-without-abstract-decorator     |
| B028 | no-explicit-stacklevel                      |
| B034 | re-sub-positional-args                      |
| B904 | raise-without-from-inside-except            |
| B905 | zip-without-explicit-strict                 |
| B911 | batched-without-explicit-strict             |
| B912 | map-without-explicit-strict                 |

Preview-only, so selecting `B` still does not turn them on: B043, B901, B903, B909.

**`UP` (pyupgrade): 48 rules, 42 on by default, 6 not.**

Stable gains: UP013 (convert-typed-dict-functional-to-class), UP015 (redundant-open-modes), UP042 (replace-str-enum).
Preview-only: UP051. Removed from ruff entirely: UP027 (0.8.0), UP038 (0.13.0).

henryiii: IMO, there are still good codes not enabled by bugbear, but pyupgrade is pretty well covered. Not enough that I'd drop UP, but not enough for me to recommend it as a dedicated check anymore, I think.

cbrnr: I agree. Feel free to add them back in another PR.

PR #325, opened 2026-08-05
Build arm64 and x86_64 wheels on macos-15

DimitriPapadopoulos:
Fix 84becab to build both arm64 and x86_64 wheels, not only arm wheels.

Really does what #280 was supposed to do.


cbrnr:
The question is if we still need macOS Intel support. It will be gone in macOS 27 anyways, and so far I haven't had any requests for this deprecated platform.

PR #323, opened 2026-08-05
Update installation instructions from pip to uv

DimitriPapadopoulos:
I have applied distinct commits for:
1. `pip` → `uv pip`
2. `uv pip` → `uv`

Fixes #260.

Issue #322, opened 2026-08-05
Enable pre-commit.ci?


DimitriPapadopoulos:
Would you agree to running pre-commit during CI in [pre-commit.ci](https://pre-commit.ci/) instead of a GitHub Action?

cbrnr:
I don't have a strong opinion on this. Personally, I don't really use pre-commit except for in this project, which only uses it for Ruff and Mypy. I would rather tend to move away from pre-commit and use Ruff directly. Replaying Mypy with Ty would probably be a bigger win in terms of speed, should we give it a try (last time I checked Ty wasn't ready)?

DimitriPapadopoulos:
The nice thing with pre-commit is that:
* In most cases it's able to automatically fix a PR if the developer hasn't run pre-commit (or prek) locally.
* The promise is that basic testing (mostly linting) is exactly the same on the developer's machine and in CI.

With that said, I'm not a huge fan of pre-commit either.

DimitriPapadopoulos:
I don't see ty raising issues when excluding tests and examples, as is the case for mypy:
https://github.com/cbrnr/sleepecg/blob/851528706c6857c293c9035107d711f48c85e899/.pre-commit-config.yaml#L12-L13

cbrnr:
Removed in #328.

PR #328, opened 2026-08-06
Switch to uv, remove pre-commit

### LINKED ISSUE #260: Use uv
We should switch to `uv` (docs, CI, ...).

cbrnr:
Fixes #260 and #322 and supersedes #323.


pyproject.toml
```diff
@@ -56,13 +52,17 @@ full = [  # complete package functionality
 
 [dependency-groups]
 dev = [
-    "sleepecg[docs]",
-    "sleepecg[full]",
-    "pre-commit >= 2.13.0",
+    "mypy >= 2.3.0",
```
DimitriPapadopoulos, line 55: How about moving` mypy`, `ruff` and `types-*` to a `lint` group, and include it in `dev`?

cbrnr: Why?

pyproject.toml
```diff
@@ -56,13 +52,17 @@ full = [  # complete package functionality
 
 [dependency-groups]
 dev = [
-    "sleepecg[docs]",
```
DimitriPapadopoulos, line 59: Perhaps keep `docs` in `dev`?
```diff
-    "sleepecg[docs]"
+    {include-group = "docs"},
```

pyproject.toml
DimitriPapadopoulos: I find it easier to understand the purpose of dependencies if they're split in smaller categories.

cbrnr: The big caveat is that `uv sync` by default pulls in the `dev` group, nothing else. Given that there are not very many dev dependencies in the first place, I think it is better to keep them in the default group.

DimitriPapadopoulos: What I meant is **include** the `lint` group in the `dev` group so that it gets pulled automatically.
```
    {include-group = "lint"},
```

.github/workflows/cibuildwheel.yml
```diff
@@ -21,10 +21,15 @@ jobs:
...
     steps:
     - uses: actions/checkout@v7
-    - uses: actions/setup-python@v7
-      with:
-        python-version: '3.12'
-    - uses: pre-commit/action@v3.0.1
+    - uses: astral-sh/setup-uv@v7
+    - name: Install the project
+      run: uv sync --locked
+    - name: Run ruff check
```
DimitriPapadopoulos, line 27: There's a GitHub Action to run ruff:
https://github.com/astral-sh/ruff-action

On the other hand there's no action (yet) to run ty.

cbrnr: We're invoking several `uv run` things here, so I think the current pattern is pretty efficient.

pyproject.toml
cbrnr: Ah, I see. See #330.

PR #329, opened 2026-08-06
Move docs from optional-dependencies to dependency-groups

DimitriPapadopoulos:
- Move `docs` group from `[project.optional-dependencies]` to `[dependency-groups]`
- Update `readthedocs.yml` to use `build.jobs.install` with `pip install --group docs`
- Reference `docs` group in `dev` group using `group:docs`


cbrnr:
I'm working on something similar in #328 which I will merge very soon!

